PR #4, opened 2026-07-13
Add Kover code coverage

SeCherkasov:
Adds [Kover](https://github.com/Kotlin/kotlinx-kover) so we can measure test coverage across the JVM/KMP suites. Aggregated at the root over `:shared`, `:composeApp`, `:server` and `:sync-wire`.

## Usage
- `./gradlew koverHtmlReport` → `build/reports/kover/html/index.html` (aggregate "all code")
- `./gradlew koverLog` → coverage % to console
- `:server:koverHtmlReport` etc. for a single module

## How it's wired (and why not the plugins {} DSL)
A versioned `plugins {}` entry resolves its plugin marker **even with `apply false`** (verified empirically) — the same trap that keeps `io.ktor.plugin` out of the root block. The hermetic offline Flatpak build (`-Dskerry.offlineRepo`) has no Kover in its offline repo, so any `plugins {}` reference would break it.

So Kover is loaded through the root `buildscript` classpath and applied via `pluginManager.apply`, every application guarded by `System.getProperty("skerry.offlineRepo") == null`. Offline packaging, `serverOnly` and `desktopOnly` builds never resolve or apply it; `findProject` keeps the root aggregation valid when those graphs drop modules.

## Notes
- Requires **Kover 0.9.8+** — earlier releases reject the `com.android.kotlin.multiplatform.library` target used by `:shared` / `:composeApp` (kover #747/#772).
- Generated Compose resource accessors and `AppVersion` are excluded so they don't swamp the denominator.

## Baseline (line coverage, generated excluded)
Aggregate ~41%. Core is well-covered — `sync-wire` 98%, `server` 89%, shared core 80–98% (terminal/mosh/ssh/ai 93–94%). The gap is the pure `@Composable` render layer (screens near 0%), which needs `compose.uiTest` / instrumented tests — a follow-up, not this PR.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,15 @@
+// Kover (code coverage) is loaded via the buildscript classpath instead of the plugins {} DSL, and
+// applied conditionally below. A versioned plugins {} entry is resolved even with `apply false` (the
+// same trap that keeps io.ktor.plugin out of this block — see the note there), which would make the
+// hermetic offline Flatpak build (-Dskerry.offlineRepo) fail: its offline repo doesn't carry Kover.
+// Coverage is a dev/CI concern, so it is simply absent from the offline packaging build.
+buildscript {
+    if (System.getProperty("skerry.offlineRepo") == null) {
+        repositories { gradlePluginPortal() }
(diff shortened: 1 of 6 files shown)


SeCherkasov:
Superseded by #5 — renamed to the gitflow `feature/` branch convention.

PR #20, opened 2026-07-14
Add VNC (RFB) remote-desktop connections

SeCherkasov:
Adds VNC as a connection type next to SSH/Mosh/Telnet/Serial. A VNC host opens a remote-desktop tab instead of a terminal, on both desktop and Android.

What's inside:

- Our own RFB client: a pure protocol state machine in `commonMain` (`RfbCodec`) over a TCP socket (`VncTcpTransport` in `jvmSharedMain`). Encodings: Raw, CopyRect, Hextile, ZRLE, Tight (with JPEG), DesktopSize. Auth: None and VNC-Auth (DES), plus "ask every time". Pointer, keyboard (X11 keysyms) and clipboard in both directions.
- The socket is treated as untrusted: server-supplied lengths and dimensions are capped before any allocation, zlib output is bounded, and any decode error ends the session instead of crashing the app. Wire text is decoded as Latin-1, per the RFB spec.
- The remote cursor is drawn client-side (Cursor pseudo-encoding), and the local pointer is hidden over the framebuffer — no double cursor, no lag on slow links.
- Resize-to-window (SetDesktopSize) with a per-host toggle, a quality menu (Auto/Low/Medium/High), and view-only mode. Mobile adds pinch-zoom, pan and an IME keyboard bridge.
- UI extras: each host row shows its connection type's icon, and the VNC view gets a slide-over hosts drawer. All VNC screen strings are localized (en/ru/zh).
- Compose Hot Reload is back for desktop development (`./gradlew :composeApp:hotRunDesktop`), pinned to 1.1.1 for CMP 1.9.x and excluded from the offline Flatpak build.


SeCherkasov:
### Code review

**Update: fixed in [47ed1af](https://github.com/SeCherkasov/SkerrySSH/commit/47ed1af) — `RfbCodec.close()` now releases the ZRLE/Tight inflaters and is called from `closeSocket()` on every teardown path, with tests.** The same commit also addressed the lower-confidence findings from this review (key-auth state leaking into a VNC form, unreachable `cleanExit`, stale doc comments), and [64d21c1](https://github.com/SeCherkasov/SkerrySSH/commit/64d21c1) localized the hardcoded VNC UI strings.

Original finding:

1. Native zlib `Inflater` instances leak on VNC session teardown. `RfbCodec` lazily creates one persistent inflater for ZRLE and up to four for Tight (both are first in `DEFAULT_ENCODINGS`, so any real server picks them), but the codec has no `close()`/dispose method — `resetTight(id)` only closes a stream mid-session on the Tight reset bit. `VncSocketSession.closeSocket()` (the only teardown path: `close()`, `onCompletion`, and the connect-failure path) closes just the socket and never releases the codec's inflaters, so every connect/disconnect cycle strands 1–5 `java.util.zip.Inflater` native allocations on GC finalization.

https://github.com/SeCherkasov/SkerrySSH/blob/5a3fad1e20cae0ba0265f50ee8a6457e750d60b8/shared/src/commonMain/kotlin/app/skerry/shared/vnc/RfbCodec.kt#L44-L48

https://github.com/SeCherkasov/SkerrySSH/blob/5a3fad1e20cae0ba0265f50ee8a6457e750d60b8/shared/src/jvmSharedMain/kotlin/app/skerry/shared/vnc/VncTcpTransport.kt#L157-L161

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #24, opened 2026-07-17
fix: prove biometric unlock works before enabling it, and say so when a phone can't do it

### LINKED ISSUE #23: Biometric unlock silently returns to password screen on some Android ROMs (e.g. Xiaomi/MIUI) — bug or intended?
Hi! First, thanks for the great app.

**Device:** Xiaomi 15 (HyperOS / MIUI), Android 15
**App version:** 0.1.5

**What happens:**
1. Enable fingerprint unlock in settings ( enrollment prompt succeeds).
2. Lock the vault / restart the app.
3. On the unlock screen, tap the fingerprint icon — the system biometric prompt appears.
4. Authenticate successfully with fingerprint (system shows the success checkmark, prompt dismisses).
5. **The app stays on the master-password screen** — it does not enter the app. No error message is shown; it just silently returns to the password form.

**Question:** is this a bug, or is this the intended fallback behavior?

**What I found reading the code (please correct me if I'm wrong):**

In `AndroidBiometricKeyStore.kt`, `onAuthenticationSucceeded` does:

```kotlin
val authedCipher = result.cryptoObject?.cipher
cont.resume(if (authedCipher != null) Auth.Success(authedCipher) else Auth.Failed)
```

Some Xiaomi/MIUI devices report biometric success **without populating `result.cryptoObject`** (a known quirk of these ROMs). So `Auth.Failed` is returned even though the fingerprint was correct.

Then in `VaultGateController.unlockWithBiometric()`:

```kotlin
// Cancelled / Failed / Unavailable / NotEnabled — stay on the password form silently.
else -> Unit
```

So the user is left on the password screen with no feedback at all — it looks like the fingerprint "did nothing".

**Two possible issues:**
1. **`cryptoObject` null on MIUI**: would it be acceptable to fall back to the cipher passed into `authenticate()` when the system reports success? The cipher is already bound to the prompt via `CryptoObject(cipher)`, so a successful auth should authorize it. (I tested this locally and it unlocks correctly on the Xiaomi 15.)
2. **Silent failure UX**: even if the fallback above isn't desirable, a failed biometric attempt currently gives zero feedback. Would you consider showing a small "biometric unlock failed, use password" hint so users aren't confused?

Happy to open a PR for either/both if you agree with the direction. Thanks!

SeCherkasov:
Fixes #23.

## The problem, as the user sees it

You turn on fingerprint unlock. It says it's on. Next time you open Skerry you touch the sensor, the phone shows the green checkmark — and the vault stays shut. No message, nothing to do about it, back to typing the master password.

## Why it happens

Skerry doesn't use your fingerprint as a "yes, it's you" checkbox. The key to your vault is locked inside the phone's security chip, and it is that chip which requires a fingerprint before it will hand the key over. That is the whole point: even someone with full access to the app's files can't read the vault without your finger on the sensor.

On some phones — Xiaomi/HyperOS in this report — the chip accepts the fingerprint and then refuses to actually use the key. The same bug is filed against KeePassDX (#2298) and Bitwarden (#6550), so it isn't specific to Skerry.

The real mistake on our side was this: when you enabled fingerprint unlock, Skerry only checked that the chip could *lock* the key away. Whether it would ever *unlock* it again was discovered later, on the next app start, when it was too late to warn you or do anything about it.

## What changes

**Turning it on now proves it works.** Skerry locks the key, immediately unlocks it again and checks it got the same key back — before it saves anything. This costs one extra fingerprint touch during setup, and it's the touch that matters, because it's exactly what will happen every time you unlock later.

**If the strongest setup isn't accepted, weaker ones are tried.** Phones differ in what their security chip will do. Skerry now asks for the strongest arrangement first and steps down if the phone won't honour it. What never changes is the part that protects you: the key is only ever released after a live fingerprint check. If a step down means the key is kept a little less strictly, the settings screen tells you so instead of quietly accepting it.

**If nothing works, fingerprint unlock is switched off rather than pretended.** The setting stays visible, but it is off and explains why, with a "Check again" button — a system update or a reboot can fix the chip, and then it will work. Your master password keeps working the whole time; it was always the real key to the vault.

**Failures are no longer silent.** If unlock doesn't work you get "Biometrics didn't work — enter your master password", with a separate message when the sensor is temporarily locked after too many attempts. The wording is deliberately short: it shows up in one line above the password field. Dismissing the prompt yourself stays quiet, as it should.

**One bad attempt doesn't cost you anything.** A timeout, a sensor hiccup, or the app being restarted mid-setup no longer counts as "this phone can't do it". Only repeated, specific refusals from the security chip do — and even then, cancelling a prompt never deletes a fingerprint setup that already worked.

## What this deliberately does not do

There is an easier fix, used by some apps: treat the fingerprint as a simple identity check and keep the vault key readable without it. That works on every phone, including the broken ones — because the key is no longer protected by the fingerprint at all. For an app holding SSH keys and server passwords, that is a hole rather than a slightly weaker setting, so it is off the table. This is discussed in #23.

Also worth noting for anyone who followed the earlier commits here: an intermediate attempt (a key valid for a few seconds after any device authentication) was reverted. It didn't help on the reporter's phone, and it was weaker — on older Android versions a PIN would have been enough to release the key. The stored format is unchanged, so nobody has to re-enroll because of this PR.

## For reviewers

- `VaultBiometrics.enable` performs a wrap → unwrap → constant-time compare before writing `vault.bio`, over a ladder of `KeyGenParameterSpec` configurations: StrongBox + `setUnlockedDeviceRequired` → TEE + `setUnlockedDeviceRequired` → TEE. `setUserAuthenticationRequired(true)` + `AUTH_BIOMETRIC_STRONG` + a per-operation `CryptoObject` is required on every rung and never traded away.
- Only evidence about the key configuration steps down a rung: the enclave refusing an authorized operation, a key invalidated on the spot, a configuration the platform won't create, or a mismatching round trip. Cancellation, sensor lockout and generic hardware failures stop the ladder — on Android that bucket also holds timeouts and a destroyed Activity, which say nothing about the device.
- A new `BiometricResult.Unusable` distinguishes "auth succeeded, the enclave refused the operation" from an ordinary failure. At unlock, two consecutive refusals disable biometrics and persist the verdict (`vault.bio.unsupported`, scoped to the device); a single one reads as an ordinary failure, since the same signal fires on a one-off keystore glitch and the reaction is destructive.
- Keystore entries, `vault.bio` and the verdict file are written on an injected dispatcher, not from click handlers.
- `FileBioArtifactStore` and `FileBiometricSupportStore` share a small `JsonFileStore` (atomic write, hardened tmp, non-throwing read).

## Testing

`./gradlew test` is green and `:androidApp:assembleDebug` builds. New tests cover the ladder walk, the round-trip check and a round trip that comes back different, the "no rung works" verdict, refusal streaks at unlock and their reset, remembering which rung the enrollment landed on, and the cases where a failed or cancelled re-enable must leave an existing setup intact.

The one thing tests can't cover is the phone's security chip itself. That needs a device — @onepve, if you have the time, another round on your Xiaomi 15 would settle it. Either it now works, or it tells you plainly that your phone can't do it and leaves you on the master password. Both are acceptable outcomes; silently failing was not.


onepve:
Thanks for the quick update! I tested the latest version (commit `f09a070` — the time-bound key approach) on a Xiaomi 15 running HyperOS. Here is what I did:

1. Completely uninstalled the old app.
2. Fresh-installed the APK built from this PR branch (with only `applicationId` changed to `app.skerry.onepve` to avoid overwriting the release build).
3. Set a master password, then enrolled a fingerprint.
4. Selected fingerprint unlock.

**Result:** It still shows the same error: "Biometric unlock failed — please enter master password."

It does _not_ crash — the failure surface you added is working. But the biometric key is never accepted. CryptoObject or time‑bound, HyperOS seems to reject it regardless. Let me know if you would like me to capture any logs or run a debug build with extra instrumentation.

PR #29, opened 2026-07-18
fix: treat the vault password as the single account password (issue #28)

### LINKED ISSUE #28: Password design clarification: one master password or two separate passwords?
### Summary

This is a design question about the password model in SkerrySSH, because it directly affects what l10n copy I should use for the Chinese UI.

---

### Two different workflows, two different results

**Workflow A — sync password on first launch (seems reasonable):**

1. Launch app → "Create vault" → enter **vault password A**
2. App asks "Local" or "Sync" → choose **Sync**
3. Enter **email** + **password B** (for sync account)
4. Enter the app normally
5. App unlock → uses **password A**
6. Sync → uses **password B**

Result: user assumes vault password ≠ sync password. Two passwords, two purposes. 🟢 No surprises.

**Workflow B — add sync later from local mode (the trap):**

1. Launch app → "Create vault" → enter **vault password A**
2. App asks "Local" or "Sync" → choose **Local**
3. Enter the app in local-only mode
4. Go to settings → tap "Sync" → enter **email** + **password B**
5. Sync succeeds silently
6. **Next app unlock: password A no longer works. It changed to password B.** 💥

The user never expected their vault password to change. They entered B for "sync", not to overwrite their vault unlock. This feels like a bug to the user.

---

### What the code says

In `FileVault.kt:adoptDataKey()`, when the account dataKey differs from the local one, the vault metadata is rewritten with the password passed in — effectively changing the vault password.

In `SyncCoordinator.kt`, the password entered on the sync screen is used for **everything**: Argon2id master key, SRP auth key, dataKey wrapping key. It is clearly a **single master password** design.

But the UI asks for it in two places with different context (vault creation → sync setup), and there is no hint they should be the same. In Workflow B, the user thinks they are adding a separate sync password — not overwriting their vault password.

---

### What I am asking

Is the intended design:

**A) One master password** — vault password and sync password are the **same thing**. The sync screen should NOT ask for a "new" password separate from the vault password; it should either reuse it or explicitly warn the user that this password will replace the vault password.

**B) Two separate passwords** — vault password protects local data, sync account password authenticates to the server. They can differ. This would require `adoptAccountDataKey` to NOT overwrite the vault password.

I need to know which one it is so I can translate the UI strings correctly in Chinese. Right now calling it "同步账户主密码" (sync account master password) implies it is separate from the vault password, which is misleading if the answer is (A).

Thank you!

SeCherkasov:
## Problem

Issue #28: enabling sync could silently change the local vault unlock password, or silently
leave the local vault password and the account password diverged. Root cause: the sync flow
treated the "sync password" as an independent secret and used it for the account while the
vault kept its own password.

The account (remote) password is the **single source of truth** — every device shares one
account dataKey wrapped under it, so a synced vault must unlock with the account password. And
the account password is immutable via connect (SRP login requires the correct password; the
server has no verifier-update endpoint).

## Change

`SyncCoordinator.doConnect` now branches on `matchesVault = vault.verifyPassword(typed)`:

- **typed == vault password** → register a new account under it, or on CONFLICT log into our own
  account and adopt its key. Adopting here never changes the unlock password.
- **typed ≠ vault password** → never register (creating an account under a non-vault password is
  what split the local unlock password from the account password). Login-only; on success pause
  with a new `SyncStatus.NeedsPasswordReplaceConfirm` so the user explicitly confirms that this
  device will start unlocking with the account password (`confirmPasswordReplace` /
  `cancelPasswordReplace`).
- wrong / non-existent → `Unauthorized`; nothing is created.

UI: shared `PasswordReplaceConfirm` component (+ desktop modal), wired into the desktop
`SyncSetupDialog` and the mobile sync screen; `sync_password_replace_*` strings in en + ru + zh; all
exhaustive `when(status)` sites updated.

### Dismissing the confirmation (869f717)

Pausing the connect means holding the typed password until the user answers, so every way out of the
prompt has to answer it. Dismissing without answering used to leave the pending replace behind: the
password copy stayed in memory and the status was stuck on `NeedsPasswordReplaceConfirm`, i.e. a sync
indicator reading "Syncing…" with no route back to the prompt.

- desktop: `dismissPasswordReplace` declines the replace before the modal's `onDismiss` (Esc — a scrim
  click is deliberately swallowed by `ModalScrim`);
- mobile: the confirmation is inline, and the system back gesture goes through `MobileDesignApp`'s
  handler straight to `state.pop()`, bypassing the header's `onBack` — so the decline hangs off
  `onDispose` of the status branch instead, covering back, gesture and tab switch alike. Leaving via
  confirm/cancel is a no-op: both clear the pending first.

## Test plan

- [x] `SyncCoordinatorPasswordReplaceTest` (real `FileVault` + `IonspinVaultCrypto`): pause /
      confirm / cancel, happy register under the vault password, rejection of a non-vault
      password, and dismissal (status returns to `Disabled`, a stale `confirmPasswordReplace()` is a
      no-op, the vault password is untouched) — all green.
- [x] `./gradlew :composeApp:desktopTest` passes.
- [x] `./gradlew :androidApp:compileDebugKotlin` passes (desktop ⇆ Android parity).
- [x] The four flows from
      [the walkthrough on #28](https://github.com/SeCherkasov/SkerrySSH/issues/28#issuecomment-5020266781)
      run end to end against a **real server** (`server/Dockerfile` on `127.0.0.1:8080`), real `FileVault`s,
      real Argon2id and the production `KtorSyncClient` — i.e. the whole desktop stack below Compose:
  - connect with the vault's own password → `Online`, no prompt, the password still unlocks;
  - a non-vault password with no such account → `Unauthorized`, nothing registered (a later connect with
    the vault password still *registers* the account rather than colliding);
  - joining an account that exists under another password → pauses; cancel leaves the old password
    working; confirm re-keys the device, requests a biometric re-enroll, and the other device's record
    decrypts locally;
  - dismissing the prompt → status leaves `NeedsPasswordReplaceConfirm`, a stale confirm is a no-op, the
    vault password is untouched.
- [ ] **Not covered:** the Compose layer itself — a real Esc key press on the desktop window and the
      Android `onDispose` when leaving the sync screen. Both are one-line wirings onto code paths the run
      above exercises, but neither has been driven through an actual UI.

## Follow-up (separate PR)

Changing the account password itself — a dedicated Security action that re-wraps the account
dataKey under a new master, updates the server SRP verifier, and signals other devices to
re-authenticate — is intentionally **out of scope** here and tracked in #32.

Closes #28

### DIFF
--- a/composeApp/src/commonMain/composeResources/values-ru/strings_sync.xml
+++ b/composeApp/src/commonMain/composeResources/values-ru/strings_sync.xml
@@ -115,6 +115,12 @@
     <string name="sync_fail_sync">Ошибка синхронизации</string>
     <string name="sync_fail_revoke">Не удалось отозвать устройство</string>
 
+    <!-- issue #28: подключение к существующему аккаунту с другим паролём меняет пароль этого устройства -->
+    <string name="sync_password_replace_title">Использовать пароль аккаунта?</string>
+    <string name="sync_password_replace_body">Аккаунт «%1$s» защищён другим паролём. Чтобы синхронизировать, это устройство будет разблокироваться паролём аккаунта.</string>
+    <string name="sync_password_replace_confirm">Использовать пароль аккаунта</string>
+    <string name="sync_password_replace_cancel">Отмена</string>
+
     <!-- Идентификаторы аккаунта для Teams-приглашений (Settings → Sync / mobile Sync) -->
     <string name="sync_account_id_label">ID аккаунта</string>
     <string name="sync_sharing_fingerprint_label">Отпечаток ключа шеринга</string>
--- a/composeApp/src/commonMain/composeResources/values-ru/strings_sync.xml
+++ b/composeApp/src/commonMain/composeResources/values-ru/strings_sync.xml
@@ -99,7 +99,7 @@
 
     <!-- Причины SyncStatus.Failed (typed reason → текст; техническая деталь добавляется после двоеточия) -->
     <string name="sync_fail_vault_locked">Хранилище заблокировано</string>
-    <string name="sync_fail_unauthorized">Неверный мастер-пароль или аккаунт</string>
+    <string name="sync_fail_unauthorized">Неверный пароль или аккаунт. Для нового аккаунта введите пароль этого хранилища.</string>
     <string name="sync_fail_account_not_found">Аккаунт не найден</string>
     <string name="sync_fail_account_exists">Аккаунт уже существует</string>
     <string name="sync_fail_pairing_expired">Код паринга истёк</string>
--- a/composeApp/src/commonMain/composeResources/values-zh/strings_sync.xml
+++ b/composeApp/src/commonMain/composeResources/values-zh/strings_sync.xml
@@ -104,7 +104,7 @@
 
     <!-- SyncStatus.Failed 原因 -->
     <string name="sync_fail_vault_locked">保险库已锁定</string>
-    <string name="sync_fail_unauthorized">主密码或账户错误</string>
+    <string name="sync_fail_unauthorized">密码或账户错误。若要新建账户，请使用此保险库的密码。</string>
     <string name="sync_fail_account_not_found">未找到账户</string>
     <string name="sync_fail_account_exists">账户已存在</string>
     <string name="sync_fail_pairing_expired">配对码已过期</string>
--- a/composeApp/src/commonMain/composeResources/values-zh/strings_sync.xml
+++ b/composeApp/src/commonMain/composeResources/values-zh/strings_sync.xml
@@ -120,6 +120,12 @@
     <string name="sync_fail_sync">同步失败</string>
     <string name="sync_fail_revoke">无法撤销设备</string>
 
+    <!-- issue #28: 连接到使用其他密码的现有账户会将本设备改用该密码 -->
+    <string name="sync_password_replace_title">使用账户密码？</string>
+    <string name="sync_password_replace_body">账户「%1$s」使用其他密码保护。若要同步，此设备将改用您的账户密码解锁。</string>
+    <string name="sync_password_replace_confirm">使用账户密码</string>
+    <string name="sync_password_replace_cancel">取消</string>
+
     <!-- Teams 邀请的账户标识符（设置 → 同步 / 移动端同步） -->
     <string name="sync_account_id_label">账户 ID</string>
     <string name="sync_sharing_fingerprint_label">共享密钥指纹</string>
--- a/composeApp/src/commonMain/composeResources/values/strings_sync.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings_sync.xml
@@ -115,6 +115,12 @@
     <string name="sync_fail_sync">Sync failed</string>
     <string name="sync_fail_revoke">Couldn't revoke the device</string>
 
+    <!-- issue #28: joining an existing account whose password differs re-keys this device to it -->
+    <string name="sync_password_replace_title">Use your account password?</string>
+    <string name="sync_password_replace_body">The account “%1$s” is protected by a different password. To sync, this device will start unlocking with your account password.</string>
+    <string name="sync_password_replace_confirm">Use account password</string>
+    <string name="sync_password_replace_cancel">Cancel</string>
+
     <!-- Идентификаторы аккаунта для Teams-приглашений (Settings → Sync / mobile Sync) -->
     <string name="sync_account_id_label">Account ID</string>
     <string name="sync_sharing_fingerprint_label">Sharing key fingerprint</string>
--- a/composeApp/src/commonMain/composeResources/values/strings_sync.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings_sync.xml
@@ -99,7 +99,7 @@
 
     <!-- Причины SyncStatus.Failed (typed reason → текст; техническая деталь добавляется после двоеточия) -->
     <string name="sync_fail_vault_locked">Vault is locked</string>
-    <string name="sync_fail_unauthorized">Wrong master password or account</string>
+    <string name="sync_fail_unauthorized">Wrong password or account. For a new account, use this vault's password.</string>
     <string name="sync_fail_account_not_found">Account not found</string>
     <string name="sync_fail_account_exists">Account already exists</string>
     <string name="sync_fail_pairing_expired">Pairing code expired</string>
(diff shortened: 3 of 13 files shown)


SeCherkasov:
## Pre-merge QA — what still needs a human

The automated run covers everything below the Compose layer (see the test plan). What follows is what it
does **not** cover: the UI wiring, the platforms, and the subsystems that a re-key touches downstream.
Work top to bottom — section A is where the new code actually is, C is where the risk is.

### Setup

```bash
# Server. NB: `docker compose up` may fail with "all predefined address pools have been fully subnetted"
# — then run the built image on the default bridge instead:
docker build -f server/Dockerfile -t skerry-sync-qa .
docker run -d --name skerry-qa -p 127.0.0.1:8080:8080 \
  -e SKERRY_DB_URL=jdbc:sqlite:/data/skerry-sync.db -e SKERRY_JWT_SECRET=qa-secret \
  -v skerry-qa-data:/data skerry-sync-qa

# Desktop, against a THROWAWAY config dir — never the real ~/.config/skerry: case A2 permanently
# changes the vault unlock password.
XDG_CONFIG_HOME=/tmp/skerry-qa-1 ./gradlew :composeApp:run     # device 1
XDG_CONFIG_HOME=/tmp/skerry-qa-2 ./gradlew :composeApp:run     # device 2 (separate vault)
```

Two gotchas that look like bugs but aren't: the server allows **5 registrations per IP per 60 s**
(`Plugins.kt`, `RateLimits.REGISTER`) — a burst of retries gives `429`, restart the container to reset the
buckets; and the server speaks plain HTTP, so use `http://127.0.0.1:8080` verbatim.

Throughout: **A** = the password a local vault is created with, **B** = the password of an account that
already exists on the server, **C** = neither.

### A. Desktop UI — the part that has never been executed

- [ ] **A1. Vault password connects clean.** New vault under A → Local → Settings → Sync → connect with A.
      Expect: `Online`, no prompt. Lock and unlock with A.
- [ ] **A2. The consent path.** Second config dir, vault under A; the account already exists under B
      (device 1). Connect with B → the confirmation appears **instead of** connecting. Press **Cancel** →
      back to the form; lock/unlock still takes A, not B. Connect with B again → **confirm** → syncs;
      lock; now B unlocks and A does not.
- [ ] **A3. Wrong password.** Fresh vault under A, unused account id, connect with C → an
      "unauthorized"-style error. Then connect with A → it must **register** the account, not report a
      collision.
- [ ] **A4. Esc on the confirmation** ← *the line added in 869f717, never run.* Reach the confirmation,
      press **Esc**. Expect: the dialog closes **and** the sync indicator in the status bar returns to its
      normal state — it must not sit on "Syncing…". Reopen Settings → Sync: the plain form, not a stale
      confirmation. Connect with A afterwards and check it still works (nothing left pending).
- [ ] **A5. Scrim click does nothing.** On the confirmation, click the dimmed area outside the card —
      by design it is swallowed, the dialog stays. (If it closes, `ModalScrim`'s contract changed.)

### B. Android — only compile-checked so far

`ANDROID_HOME=$HOME/Android/Sdk ./gradlew :androidApp:installDebug`, server reachable from the phone
(LAN IP, not 127.0.0.1).

- [ ] **B1.** Repeat A1–A3 on the phone. The confirmation is **inline on the Sync screen**, not a modal.
- [ ] **B2. System back gesture** ← *the `onDispose` added in 869f717.* On the confirmation, leave with the
      gesture (not the header arrow). Return to Sync: the form, not the confirmation; the vault still
      unlocks with A.
- [ ] **B3. Header back arrow** — same expectation.
- [ ] **B4. Tab switch.** From the confirmation, switch to another tab and come back: form, not
      confirmation.
- [ ] **B5. Rotation / app to background and back** while the confirmation is up. Either outcome is
      acceptable (it stays, or it declines and returns to the form) — what must **not** happen is a stuck
      "Syncing…" or a silent password change. Note down which one you see.

### C. What a re-key drags with it — the real risk area

All of these ride on case A2 (confirmed replace), and none were exercised by the automated run.

- [ ] **C1. keep-connected across a restart.** Enable "keep connected" on device 2 before joining, confirm
      the replace, then quit and relaunch. Expect: it restores the session without asking for a password
      (the refresh token is re-sealed under the *new* dataKey). Related: #30.
- [ ] **C2. Biometric re-enroll.** With biometric unlock enabled on device 2, confirm the replace. Expect:
      a re-enroll prompt, and after re-enrolling, biometric unlock actually opens the vault. Android only
      (and see #23 / #24 for OEM-ROM caveats).
- [ ] **C3. Teams.** If device 2 was in a team before joining, check the team vault still opens after the
      re-key — team vaults are unlocked by a key stored in the account vault.
- [ ] **C4. A third device.** With devices 1 and 2 on the account, connect a third under B. Expect: no
      prompt if its own vault password is already B; the usual confirmation otherwise. No device loses
      access.
- [ ] **C5. Records both ways.** After A2, create a host on device 1 → it appears on device 2, and back.

### D. Wording on screen

- [ ] **D1.** The confirmation in **en / ru / zh** — read it as someone who typed the sync password
      expecting it to be separate. It has to say plainly that *this device will start unlocking with the
      account password*. Strings: `sync_password_replace_*`.
- [ ] **D2.** The failure text for case A3 points at the vault password rather than just saying "wrong
      password".

### E. One decision to make before merging

`doConnect`'s confirmed branch treats `adoptAccountDataKey` returning `false` as success: we'd connect
with the **local** key, leave the vault password unchanged, and the user would believe they consented to a
change that didn't happen — the same divergence this PR removes. It is unreachable in normal operation
(the login just succeeded with that very password), so there's no test for it. Decide: add a guard that
fails the connect on `adoptedKey == false` in that branch, or document why it can't happen.

---

Anything that fails here, note the case number — the fix belongs in this PR, not a follow-up.

SeCherkasov:
Point **E** of the QA checklist above is closed in code — it turned out to be a real, reachable hole rather than a theoretical one, and the guard I proposed there would have been the wrong fix.

`FileVault.adoptDataKey` returns `false` for two different reasons: the wrap didn't open, **or** the incoming key already equals the current one — in which case it deliberately keeps the metadata, so that a primary device reconnecting with its own key doesn't get its password changed. The second case is reachable today: register an account under a password, change the vault password locally in Settings → Security, then reconnect with the account password. The confirmation appears, the user agrees that this device will unlock with the account password — and nothing is re-wrapped, because there is no key to adopt. Vault on one password, account on another, with consent on screen.

`adoptAccountDataKey` now reports three outcomes:

- **Adopted** — as before, full re-pull;
- **AlreadyOurs** — re-wrap the metadata under the account password (new `Vault.rewrapUnder`, a default `false` in the interface so existing implementations keep compiling);
- **Undecryptable** — fail the connect instead of reporting success for a replace that didn't happen.

Two tests in `SyncCoordinatorPasswordReplaceTest`, both verified red against the previous commit before the fix went in. `./gradlew test :composeApp:desktopTest` and `:androidApp:compileDebugKotlin` are green.

This adds one manual case to section A, since it changes what the confirmed path does:

- [ ] **A6.** Vault under A, connect with A (account gets created). Settings → Security → change the vault password to X. Reconnect with A → confirmation → confirm. Expect: connects, and the vault now unlocks with **A**, not X. Before this commit it stayed on X.

PR #46, opened 2026-07-21
Built-in SSH agent with forwarding into sessions

SeCherkasov:
Keys stay in the vault; a session — or another program on this machine — only ever gets a signature.

## What this adds

- **Agent core** (`shared/commonMain/agent`): the SSH agent protocol as a pure codec. Only requests that cannot change the key set are served — listing and signing; add/remove/lock/extension get `SSH_AGENT_FAILURE`, because the key set is the vault's, not the peer's. Every field length is bounded by what actually arrived.
- **Signing** (`jvmSharedMain`): vault keys parsed with the same stack the transport authenticates with (sshj + BouncyCastle). `rsa-sha2-256/512` flags honoured; a certificate credential is offered together with the plain key behind it, as `ssh-add` does.
- **Forwarding**: sshj 0.40 has no agent support at all, so both halves are hand-built — a `SessionChannel` subclass sends `auth-agent-req@openssh.com` before the PTY, and a `ForwardedChannelOpener` answers the server's `auth-agent@openssh.com` channels. Shell only: `exec` channels (metrics, service scan, the Mosh bootstrap) never carry keys.
- **`SSH_AUTH_SOCK`** (desktop): a unix socket under `$XDG_RUNTIME_DIR`, 0700 directory and 0600 socket. Android has nothing to serve, so the row is absent there rather than faked.
- **UI**: Settings → SSH agent (master switch, which keys are in the agent, the socket, recent activity) on both platforms, plus a per-host "Agent forwarding" toggle.

## Exposure gates

Three independent ones: the agent off, the vault locked, or the key not put in the agent. Forwarding is per host and off by default — while the session lives, that host can use every key the agent offers. Sessions deliberately survive a vault lock, so the lock drops parsed keys and closes the socket. `forwardAgent` is stripped when a host is shared to a team: trusting a host with your own keys is each member's decision.

## Review

Four independent reviews (security, Kotlin/concurrency, protocol conformance vs OpenSSH, UX/i18n/parity) ran before this PR; their findings are fixed in the second commit — see its message. Known follow-ups, deliberately not in scope: per-signature confirmation (`ssh-agent -c` style), an ECDSA signing test (the fixture, not the path — sshj reads a narrow set of EC PEM formats), and a badge in the Vault section showing which keys are in the agent.

## Verification

`./gradlew build` green with lint on desktop and Android. 39 new tests, including an end-to-end run against an embedded Apache MINA SSHD server: it opens the agent channel, gets the identity list and an Ed25519 signature that verifies under the offered key; sibling tests prove a host without the flag is refused the channel, and that disconnecting ends the agent. Screens checked visually via the offscreen renderer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,8 @@ Building it yourself is also easy — see [Building from source](#building-from-
 
 ## Features
 
-- **Connections** — SSH with jump hosts (ProxyJump) and SSH certificates; SFTP (dual-pane
+- **Connections** — SSH with jump hosts (ProxyJump) and SSH certificates; built-in SSH agent
+  with per-host forwarding (and `SSH_AUTH_SOCK` for local tools on desktop); SFTP (dual-pane
   commander); port forwarding: local, remote, dynamic/SOCKS; Telnet; serial (desktop and
   Android USB-OTG).
 - **Terminal** — custom grid emulation, session tabs with split view, SSH auto-reconnect,
--- a/README.ru.md
+++ b/README.ru.md
@@ -96,7 +96,8 @@ macOS)** и **Android**, паритет фич между платформами
 
 ## Возможности
 
-- **Подключения** — SSH с jump-хостами (ProxyJump) и SSH-сертификатами; SFTP (двухпанельный
+- **Подключения** — SSH с jump-хостами (ProxyJump) и SSH-сертификатами; встроенный SSH-агент
+  с пробросом по хостам (и `SSH_AUTH_SOCK` для локальных утилит на десктопе); SFTP (двухпанельный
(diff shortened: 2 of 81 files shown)


SeCherkasov:
Closing without merging — parked, not abandoned.

The branch `feat/ssh-agent-forwarding` stays on the remote with everything in it: the agent and forwarding, per-signature confirmation (`ssh-agent -c`), PKCS#8 key reading, the deterministic socket shutdown, and the vault badge. CI was green and the forwarding path was exercised by hand against a test server.

Reopen or rebase when the feature is wanted; the review findings from this round are already fixed in the branch.

PR #47, opened 2026-07-21
Let a host use only the agent keys it needs

SeCherkasov:
Builds on #46 (base branch is `feat/ssh-agent-forwarding`, so this merges after it).

## Why

Agent forwarding hands the far side a live agent. Until now that meant **every** key in it: any host with forwarding on could ask for a signature by any key, and could see the names of all of them. Per-signature confirmation helps, but it is off by default and does not hide the listing.

## What

- A host profile carries the vault keys it may reach (`Host.agentKeyIds`). Empty means every key — what existing profiles say, and what OpenSSH does, since its agent has no per-host sets.
- The set travels with `SshTarget` into the forwarder and down to the keyring (`SshAgentScope`), which filters what comes out of its shared cache, so narrowing one host costs no other host a re-parse.
- The picker sits under the forwarding switch in the host form, not in Settings — that is where the user can see which host they are deciding for.

## Notes

- Sync-safe: `Host` is JSON with `ignoreUnknownKeys`, so an older client keeps the profile intact and simply behaves as "all keys".
- The local agent socket is unaffected — it is not a host and keeps the full set.

## Tests

`:shared:desktopTest` + `:composeApp:desktopTest` green; Android compiles. New: keyring honours a restricted set (and still serves everything to an unrestricted caller), service carries the set through, profile round-trips the selection.

Not verified by hand yet.

### DIFF
--- a/composeApp/src/commonMain/composeResources/values-ru/strings_agent.xml
+++ b/composeApp/src/commonMain/composeResources/values-ru/strings_agent.xml
@@ -37,4 +37,5 @@
     <string name="agent_prompt_hint">Без ответа в течение двух минут запрос будет отклонён.</string>
     <string name="agent_prompt_allow">Разрешить один раз</string>
     <string name="agent_prompt_deny">Отклонить</string>
+    <string name="agent_keys_desc">Это то, что держит агент. Какие из них достанутся конкретному хосту, выбирается в самом хосте.</string>
 </resources>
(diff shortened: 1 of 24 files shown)


SeCherkasov:
Closing without merging — the feature is a second permission system on top of one that already works.

Agent forwarding is off by default and enabled per host, and per-signature confirmation stops a key from being used silently. Those two already cover the risk this PR was aimed at, while a per-host key set adds a setting OpenSSH has no equivalent for, in a form where most users have one or two keys and nothing to choose between.

The branch `feat/agent-keys-per-host` stays: the shared-side mechanism (`SshAgentScope`, host profile field, keyring filtering) is small and covered by tests, so bringing it back is cheap if someone turns up with a bastion fleet and a dozen keys.

PR #51, opened 2026-07-21
Exclude revoked devices from the tombstone watermark

onepve:
Found while using the admin console on my own account: with every active device fully caught up, "Purge tombstones" still reported nothing to purge. Turned out the watermark counted **revoked** devices too — an old phone I had re-paired and revoked was pinning it forever at its frozen cursor.

**Root cause** — `tombstoneWatermark()` takes `min(lastSyncVersion)` over *all* of an account's devices. A revoked device can never pull again (the server rejects its auth), so its cursor never advances, and any tombstone above it stays unpurgeable for the account's lifetime (the team-scope age purge doesn't cover account records).

**Fix** — exclude `revoked = true` from the watermark. A revoked device carries no resurrection risk because it can never pull. If it is ever re-activated (re-authentication clears revocation per `DeviceRepository`), its next pull simply never sees tombstones purged while it was revoked — the same end state as a device that was offline past the team-scope age purge, and the record itself is long gone from every peer.

**Test** — regression test in `AdminRepositoryTest`: active device caught up + revoked device stuck at an old cursor → tombstone purges. Full `:server:test` suite green.

Real-world numbers from my account before the fix: 6 devices (2 active at cursor 59, 4 revoked at cursors 1/41/43/44), 6 tombstones at seq 6–43, watermark = 1 → 0 purgeable. After the fix: watermark = 59 → all 6 purgeable. (•̀ᴗ•́)و

### DIFF
--- a/androidApp/src/main/kotlin/app/skerry/android/AndroidSyncConfigStore.kt
+++ b/androidApp/src/main/kotlin/app/skerry/android/AndroidSyncConfigStore.kt
@@ -31,6 +31,7 @@ class AndroidSyncConfigStore(private val file: File) : SyncConfigStore {
                 deviceId = device,
                 keepConnected = map["keepConnected"] == "true",
                 sealedRefreshToken = map["sealedRefreshToken"]?.takeIf { it.isNotEmpty() },
+                pendingReconcile = map["pendingReconcile"] == "true", // absent in older files → false
             )
         }.getOrNull()
     }
@@ -42,6 +43,7 @@ class AndroidSyncConfigStore(private val file: File) : SyncConfigStore {
             appendLine("deviceId=${URLEncoder.encode(config.deviceId, "UTF-8")}")
             appendLine("keepConnected=${config.keepConnected}")
             config.sealedRefreshToken?.let { appendLine("sealedRefreshToken=${URLEncoder.encode(it, "UTF-8")}") }
+            if (config.pendingReconcile) appendLine("pendingReconcile=true")
         }
         // Atomic write: write to a temp file then rename, so a process kill mid-write can't leave a truncated sync.json.
         val tmp = File(file.parentFile, "${file.name}.tmp")
--- a/composeApp/src/commonMain/kotlin/app/skerry/ui/sync/SyncCoordinator.kt
+++ b/composeApp/src/commonMain/kotlin/app/skerry/ui/sync/SyncCoordinator.kt
@@ -55,6 +55,14 @@ data class SyncConfig(
     val deviceId: String,
     val keepConnected: Boolean = false,
     val sealedRefreshToken: String? = null,
+    /**
+     * Durable "this device must rebuild from the server before pushing" marker for the revoked→reactivated
+     * flow. The server clears revocation on the SRP verify that reports `reactivated`, so it never reports
+     * it again; persisting the intent here (set before the vault is cleared, cleared only after the first
+     * sync succeeds) means an interrupted reconcile is retried on the next connect/restore instead of
+     * silently resurrecting a purged record. Default `false`; older config files without the key load as `false`.
+     */
+    val pendingReconcile: Boolean = false,
 )
 
 class InMemorySyncConfigStore : SyncConfigStore {
@@ -487,12 +495,17 @@ class SyncCoordinator(
             // (onVaultReset) and adoptedKey.
             // A reactivated (formerly revoked) device forces a full re-pull like an adopted key, and first
             // discards its pre-revocation records so a stale live copy can't re-push a server-purged record.
+            // The server reports `reactivated` only once (it clears revocation on this verify), so we also
+            // honor a durable pendingReconcile from a previously interrupted reconcile — otherwise a crash
+            // mid-reconcile would drop the signal and let the stale records push on the next connect.
+            val pendingReconcile = configStore.load()?.takeIf { it.accountId == accountId }?.pendingReconcile == true
+            val mustReconcile = reactivated || pendingReconcile
             activateSession(
                 syncClient,
                 newSession,
                 SyncConfig(serverUrl, accountId, deviceId, keepConnected, sealed),
-                resetCursor = adoptedKey || reactivated,
-                clearLocalRecords = reactivated,
+                resetCursor = adoptedKey || mustReconcile,
+                clearLocalRecords = mustReconcile,
             )
         } catch (e: CancellationException) {
             throw e // don't swallow cancellation — it would break structured concurrency
@@ -533,18 +546,37 @@ class SyncCoordinator(
         val superseded = client?.takeIf { it !== syncClient }
         client = syncClient
         session = newSession
-        // Reactivation reconcile: drop the sync-eligible local records BEFORE resetting the cursor and
-        // running the first sync, so the following full re-pull rebuilds them from the server snapshot and
-        // the subsequent push can't resurrect a record the server purged while this device was revoked.
-        // Scoped to currently-synced types (what the push would send), so device-local data survives.
+        // Reactivation reconcile: drop the local records BEFORE resetting the cursor and running the first
+        // sync, so the following full re-pull rebuilds them from the server snapshot and the subsequent
+        // push can't resurrect a record the server purged while this device was revoked.
+        //
+        // The drop set is EVERY sync-capable type, NOT the currently-enabled ones: the push after the
+        // reconciling pull filters by the SERVER's "what syncs" settings (the pull applies them), which can
+        // differ from this device's stale local settings. If we gated the clear by the local settings, a
+        // type disabled locally but enabled on the server would keep its stale record through the clear and
+        // then be pushed once the pull flips the filter on — resurrecting the purged record. Clearing by the
+        // maximal (everything-on) filter covers any server settings; only never-synced, device-local types
+        // (terminal history) are kept.
         if (clearLocalRecords) {
-            val filter = settingsStore.load()
-            vault.clearRecords(RecordType.entries.filter { filter.shouldSync(it) }.toSet())
+            // Persist the reconcile intent BEFORE mutating the vault: the server clears revocation on the
+            // verify that reported `reactivated` and never reports it again, so a crash between here and the
+            // first successful sync would otherwise lose the signal and let the stale records push. The
+            // marker is cleared only after runSync succeeds below, so an interrupted reconcile is retried.
+            configStore.save(config.copy(pendingReconcile = true))
+            val syncCapable = SyncSettings(syncHosts = true, syncSnippets = true)
+            vault.clearRecords(RecordType.entries.filter { syncCapable.shouldSync(it) }.toSet())
+            syncState.setCursor(config.accountId, 0) // reactivation always full-pulls to rebuild from the server
+        } else {
+            if (resetCursor) syncState.setCursor(config.accountId, 0)
+            configStore.save(config)
         }
-        if (resetCursor) syncState.setCursor(config.accountId, 0)
-        configStore.save(config)
         health.setTarget(config.serverUrl)
         runSync()
+        // The reconcile is complete only once the first full re-pull actually succeeded (status Online);
+        // until then keep the marker so an interrupted or failed reconcile is redone on the next connect.
+        if (clearLocalRecords && _status.value is SyncStatus.Online) {
+            configStore.save(config.copy(pendingReconcile = false))
+        }
         startWatch()
         startLocalPush()
         // Reconnecting over a live session (switching accounts, a confirmed password replace) leaves the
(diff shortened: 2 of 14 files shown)


SeCherkasov (commented):
The bug is real and the fix targets it correctly, but the resurrection-safety argument has a gap I'd like to resolve before this lands.

**Real bug — confirmed.** A revoked device is rejected on auth, so its `lastSyncVersion` never advances and, taken into `min(...)`, pins the watermark forever. One revoked-and-stuck device makes account-scope purge a permanent no-op. Agreed.

**Gap — a re-activated device can still resurrect a purged record via *push*, not pull.** The doc comment reasons only about the reactivated device's next *delta pull* ("never sees a deletion for a record it may still hold"). But sync is bidirectional and the client re-pushes its **entire** local set every cycle:

- `SyncEngine.sync()` pushes all local records, not just changed ones:
  ```kotlin
  val local = vault.records().filter { filter.shouldSync(it.type) }.map { it.toRemote() }
  if (local.isNotEmpty()) client.push(session, local)
  ```
- Re-authentication of a revoked device clears `revoked` via `DeviceRepository.register()` but does **not** reset the client cursor or wipe the local vault: `SyncCoordinator` resets the cursor to 0 only when the dataKey changed (`resetCursor = adoptedKey`). Same-password re-auth ⇒ `adoptedKey = false` ⇒ incremental resume with the stale local vault intact.

Repro:
1. `devOld` synced to cursor 1, still holds live record `r1` locally.
2. `r1` is deleted on another device → tombstone.
3. `devOld` is revoked from the admin console before it ever pulls that tombstone.
4. Admin purges → with this change the watermark is the active cursor, so `r1`'s tombstone row is physically deleted.
5. User re-enters the master password on `devOld` (supported path — `register()` clears revocation, cursor not reset).
6. `devOld` syncs: the pull from cursor 1 never delivers the purged tombstone; the push then re-sends `r1` v1 → server `existing == null` → **`r1` is resurrected as a live record** and propagates back to the active device.

That's exactly the resurrection the watermark exists to prevent — this change removes the guard for revoked devices, and the reactivation path is officially supported, so the window is reachable.

**Options to close it (any one):**
- On clearing `revoked` in `register()`, force a full re-pull / cursor reset for that device so it can't push stale local state; or
- Server-side, prevent a push from resurrecting an id whose tombstone was purged (a purge blocklist / minimum-version guard); or
- At minimum, narrow the window and document it — e.g. purge only tombstones older than an age threshold, mirroring the team-scope age purge this PR references.

The watermark bug itself is worth fixing — just want the reactivation resurrection handled or explicitly bounded rather than relocated.

SeCherkasov:
Thanks for the solid find and the detailed write-up. The watermark bug is worth fixing, so I'd like to get this in — but let's close the reactivation resurrection window I described above first. I'll hold off on merging and wait for your changes. Happy to discuss any of the three options if one is easier on your end.

onepve:
Great catch — the push-path resurrection via reactivation is real, and the repro steps make the window concrete. Agreed this needs to be closed rather than relocated.

Of the three options, I'd go with **option 1 (full re-pull on reactivation)** — it matches the semantics we already have for vault reset (`adoptedKey = true` → `resetCursor = true`): "this device's local state can't be trusted, start over". The reactivated device then physically cannot hold records newer than the server, so both the pull gap *and* the push resurrection are closed by the same mechanism. Option 2 (server-side push blocklist) adds a second persistent state to maintain and expire; option 3 only narrows the window.

Implementation-wise, the cleanest signal I see is server-driven: when `DeviceRepository.register()` clears `revoked`, the login/register response carries a `reactivated: true` flag, and `SyncCoordinator` treats it like `adoptedKey` — cursor reset + full pull. That keeps the truth on the server side (works across reinstalls and same-password re-auth, where the client can't reliably know it was revoked). The alternative is a purely client-side marker ("last sync failure was auth-revoked"), but that misses reinstalls and feels more fragile.

Are you OK with the response-schema addition (`reactivated` on the auth response)? If you'd rather keep the wire format untouched, I can do the client-side marker instead — just say which you'd prefer and I'll push the change.

SeCherkasov:
Here's the concrete, end-to-end plan we'd like this to follow. It closes the resurrection window at its root — restoring the invariant the watermark relies on ("once a tombstone is purged, no device still holds the record *live*") — rather than narrowing it.

### The invariant, and why cursor-reset alone doesn't restore it

The watermark is safe because a tombstone is only purged once **every** device has pulled it, so every device holds the record as a *tombstone*, not live — and `compact` then drops tombstones and stops pushing them. Excluding revoked devices breaks that: a revoked device can still hold the record **live**. On reactivation it re-pushes it (`SyncEngine` pushes `vault.records()` in full), and `mergeRemote` never deletes a local record the server no longer has, so a cursor reset + full pull leaves the stale live record in the vault → push resurrects it.

So the reactivated device must be made to **discard its pre-revocation local records and rebuild from the server snapshot**. Defined semantics: *revoke + reactivate ⇒ the device becomes a fresh mirror of the server; any offline-only edits it made while revoked are intentionally dropped* (they were never pushable anyway — auth was rejected). Document this as the intended contract.

### Implementation

**1. Server — surface the signal you already compute (minimal wire change).**
`AuthRoutes` already has `val reactivated = services.devices.register(...)` at the SRP-verify path (line 112) and logs `device.reenrolled`. Just carry it to the client:
- `sync-wire/SyncWireDto.kt`: `data class VerifyResponse(val m2, val accessToken, val refreshToken, val reactivated: Boolean = false)` — the default keeps old clients/servers wire-compatible.
- `AuthRoutes`: `VerifyResponse(..., reactivated = reactivated)`.
- SRP verify is the only reactivation choke point that keeps an existing vault: account-create is a new account, pairing-claim already rebuilds the vault fresh, and `/auth/refresh` never calls `register()` (a revoked device can't refresh — `isRevoked` rejects its JWT). So `VerifyResponse` is the only response that needs the flag. (Optional: echo it on `PairingClaimResponse` too for symmetry, but it's already safe there.)

**2. Vault — add a scoped "drop records, keep the key" primitive (new; `reset()` is the wrong tool).**
`Vault.reset()` nukes the whole file including the dataKey — that's the forgotten-password exit, not this. Add a soft reset:
- `Vault`: `fun clearRecords() {}` (default no-op for fakes).
- `FileVault.clearRecords()`: under `lock`, `commit(session(), emptyList())` — empties the record list, keeps salt/wrappedDataKey/unlock state, and **does not** emit `localChanges` and does **not** create tombstones (unlike `remove()`, which tombstones and would re-inflate the very tombstones we purged).

**3. Client — reconcile on reactivation, before the first push.**
- Thread `reactivated` out of the SRP login in `KtorSyncClient` into whatever it returns to `SyncCoordinator.doConnect`.
- `SyncCoordinator.activateSession(...)`: add a `clearLocalRecords: Boolean` param. Body, in order: `if (clearLocalRecords) vault.clearRecords()` → `if (resetCursor) syncState.setCursor(0)` → `runSync()`. Ordering matters: `clearRecords` must run before `runSync()`/`startLocalPush()` so the first push sees the server-mirrored (post-pull) vault, not the stale one.
- In `doConnect`, when `reactivated` is true pass `clearLocalRecords = true` and force `resetCursor = true` (full pull to repopulate). After clear+pull the vault equals the server snapshot; the subsequent push is an idempotent no-op (same versions lose LWW), and the purged record never comes back.

### Tests (TDD — write these first)

- **Client regression (the load-bearing one)** — `SyncEngine`/coordinator level via the existing `FakeSyncClient`/`SyncE2eTest` harness: vault holds live `r1`; server has no row for `r1` (purged); trigger the reactivation path → assert (a) `r1` is gone from the local vault and (b) the server still has no `r1` after the sync (no push win). **Negative check**: revert the `clearRecords` call and this test must fail (r1 resurrected) — if it stays green both ways it's asserting nothing.
- **Vault** — `clearRecords()` empties records, keeps the vault unlocked with the same dataKey, and emits no `localChanges`.
- **Server** — SRP-verify returns `reactivated = true` for a revoked device re-logging in (and `false` normally). Keep the existing `AdminRepositoryTest` watermark test as-is.

### One thing to verify while implementing

If team-scope records share the same `Vault` as account records, `clearRecords()` would drop team records too, and the account full-pull won't restore them. Either scope `clearRecords()` to account records, or have the reactivation reconcile re-pull team records as well. Please confirm the vault's account/team record layout before wiring the clear.

### Not needed

We considered a server-side push blocklist / min-version guard, but any correct version needs unbounded per-id memory of purged records — which defeats purging. And a "reject pushes while cursor < watermark" backstop doesn't work because `SyncEngine` pulls before it pushes, advancing the cursor past the watermark first. The client reconcile above is the correct, bounded place for the guard. Age-thresholded purge (mirroring team-scope) is fine as optional extra hardening but isn't required once the reconcile is in.

Keep the `revoked eq false` watermark change from this PR — that part is correct. This plan is the second half that makes it safe. Happy to review in slices (wire + server first, then vault primitive, then the client reconcile + tests) if that's easier to land.

onepve:
Thanks for the detailed plan — the analysis is impressive and I learned a lot from it.

Honestly though, I only have some basic programming knowledge, and this fix is a bit too difficult for me right now. I'm not confident I could implement it correctly 😅

Please feel free to take over the implementation if you'd like — and no need to close this PR, I'll leave it open in case I (or anyone) can pick it up later.

For now I need to focus on adapting the app for China's domestic OS platforms (Kylin, UOS, etc.) — my workplace leadership now requires us to use these systems, so I have to prioritize that for work. I'll come back to this when I have more time and better skills.

Thanks again for the thorough review!

SeCherkasov:
No worries at all — thanks for the great find and for leaving the PR open. I've pushed the second half onto your branch (`63abdaf`), implementing the plan end to end so this can land safely:

- **Server**: `VerifyResponse` now carries `reactivated` (default `false`, wire-compatible), set from the value `DeviceRepository.register` already computes.
- **Wire → client**: the flag rides `SyncSession` out of the SRP login into `SyncCoordinator`.
- **Vault**: new `Vault.clearRecords(types)` drops records of the given types with **no tombstone** and **no push signal**, keeping the key/metadata/unlock state (`FileVault` implements it).
- **Coordinator**: on a reactivated login it clears the sync-eligible records and forces a full re-pull *before* the first push, so a purged record can't be resurrected.

One deliberate refinement over the sketch: `clearRecords` is **scoped to the currently sync-eligible types** rather than wiping everything, so genuinely device-local data (terminal history, or a type you've turned off in "What syncs") isn't lost on reactivation. Since only sync-eligible types are ever pushed, this still closes the resurrection window completely.

Your `revoked eq false` watermark change is kept as-is. Tests added: server (`reactivated` true on revoked re-login, false normally), `FileVault.clearRecords`, and a coordinator regression that drives the reactivation path with a real vault + engine and asserts the stale record is dropped locally and never re-pushed (verified it fails if the reconcile is removed). Full `:server:test`, shared sync/vault, and coordinator suites are green.

onepve:
Wow, thank you so much for implementing the whole thing yourself! (≧▽≦) You didn't have to do that — I really appreciate it ヽ(✿ﾟ▽ﾟ)ノ

And honestly, thank you for building Skerry in the first place. I searched for a long, long time before finding an SSH client that's this clean, efficient, and beautiful. The UI is gorgeous, the sync just works, and the whole thing feels so polished. It's rare to find software this well-crafted these days (´▽`ʃ♡ƪ)

The scoped `clearRecords` refinement is a nice touch too — way better than the blunt "wipe everything" approach I would have ended up with lol

Thanks again for the amazing work! ☆⌒(≧▽°)

SeCherkasov:
Ran the diff through a Kotlin/correctness and a security review pass before merge and pushed the fixes as `e6d2e25`:

- **Settings-drift resurrection (med-high)**: the reconcile's `clearRecords` was scoped by this device's *local* "what syncs" settings, read before the reconciling pull — but the push after the pull uses the *server's* settings. A type disabled locally but enabled on the server kept its stale record through the clear and got pushed once the pull flipped the filter on. Now cleared by the maximal (everything-on) filter, so it's independent of the (stale) local toggles; only never-synced device-local data (terminal history) survives.
- **Non-durable reactivation signal (high)**: the server clears revocation on the verify that reports `reactivated`, so it's a one-shot — a crash between login and the reconcile finishing lost the signal and let the stale records push on the next connect. Now persisted as `SyncConfig.pendingReconcile` (set before the vault is cleared, cleared only after the first re-pull succeeds, honored on every connect and on keep-connected restore), so an interrupted reconcile is retried.

Confirmed clean: `reactivated` is server-computed after a successful SRP verify (not client-forgeable), revocation genuinely blocks all authenticated sync endpoints (JWT validator re-checks `isRevoked` per request), and key material handling is untouched. Added regressions for both fixes (incl. a negative check). All suites green.

SeCherkasov:
Honestly, all of this was built first and foremost for ourselves, and with a lot of love — so digging into this edge case wasn't a chore, it was a pleasure. Thank you for the warm words and the kind gratitude; they mean a lot.

And best of luck with the Kylin/UOS work — no rush at all coming back here.

If your company ends up using this software, I'll be genuinely looking forward to an invitation to your corporate party 🎉

PR #52, opened 2026-07-22
feat(desktop): optional double-click to connect to a host

onepve:
Hi! 😊

Host rows in the terminal sidebar connect on a **single click** today. When you're just browsing the catalog it's easy to accidentally open a session on a production machine — one stray click and you're connected.

This PR adds **Settings → Terminal → Behavior → "Connect to host on"** with two options:

- **Single click** (default) — the current behavior, unchanged
- **Double click** — the file-manager convention: click selects, double-click opens

Details:

- `HostClickConnectMode` persisted to prefs; the default keeps single click so existing users see no behavior change
- The three host-row types in `HostsSidebar` (team / recent / catalog) switch between `clickable` and `combinedClickable(onClick={}, onDoubleClick=...)` via a small `Modifier.hostConnectClick` helper
- Mobile is unaffected (always connects on tap)
- i18n: en / zh / ru

Verified with `:composeApp:compileKotlinDesktop`.

### DIFF
--- a/composeApp/src/commonMain/composeResources/values-ru/strings_settings.xml
+++ b/composeApp/src/commonMain/composeResources/values-ru/strings_settings.xml
@@ -87,6 +87,9 @@
     <string name="settings_terminal_show_title">Показывать заголовок терминала на вкладках</string>
     <string name="settings_terminal_show_title_desc">Разрешить подключённому шеллу переименовывать вкладку (например, по запущенной программе). Выкл — вкладки всегда показывают сохранённое имя хоста.</string>
     <string name="settings_terminal_clipboard_write">Разрешить серверу писать в буфер обмена (OSC 52)</string>
+    <string name="settings_terminal_host_connect">Подключение к хосту</string>
+    <string name="settings_terminal_host_connect_single">Один клик</string>
+    <string name="settings_terminal_host_connect_double">Двойной клик</string>
     <string name="settings_terminal_clipboard_write_desc">Разрешить подключённому шеллу заменять системный буфер обмена (используется copy в tmux/vim). По умолчанию выкл — иначе вредоносный хост может подсунуть свой текст в то, что вы вставите.</string>
 
     <!-- Раздел Sync (включая карточку аккаунта и устройства — бывшая вкладка Account влита сюда) -->
--- a/composeApp/src/commonMain/composeResources/values-ru/strings_settings.xml
+++ b/composeApp/src/commonMain/composeResources/values-ru/strings_settings.xml
@@ -89,7 +89,7 @@
     <string name="settings_terminal_clipboard_write">Разрешить серверу писать в буфер обмена (OSC 52)</string>
     <string name="settings_terminal_host_connect">Подключение к хосту</string>
     <string name="settings_terminal_host_connect_single">Один клик</string>
-    <string name="settings_terminal_host_connect_double">Двойной клик</string>
+    <string name="settings_terminal_host_connect_double">Двойной клик (клик выделяет)</string>
     <string name="settings_terminal_clipboard_write_desc">Разрешить подключённому шеллу заменять системный буфер обмена (используется copy в tmux/vim). По умолчанию выкл — иначе вредоносный хост может подсунуть свой текст в то, что вы вставите.</string>
 
     <!-- Раздел Sync (включая карточку аккаунта и устройства — бывшая вкладка Account влита сюда) -->
--- a/composeApp/src/commonMain/composeResources/values-zh/strings_settings.xml
+++ b/composeApp/src/commonMain/composeResources/values-zh/strings_settings.xml
@@ -87,6 +87,9 @@
     <string name="settings_terminal_show_title">在标签页上显示终端标题</string>
     <string name="settings_terminal_show_title_desc">允许连接的终端程序重命名标签页（如运行中的程序名）。关闭——标签页始终显示保存的主机名。</string>
     <string name="settings_terminal_clipboard_write">允许服务器写入剪贴板 (OSC 52)</string>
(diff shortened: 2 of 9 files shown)


SeCherkasov (commented):
This one is in good shape — it follows the existing settings plumbing correctly (id-backed enum + `prefs.id`, `DesktopDesignState` field + `chooseHostClickConnectMode`, `staticCompositionLocalOf`, i18n en/ru/zh), and the three connect paths it gates are the right ones: `TeamHostRow` and `RecentHostRow` (`connect(host)`) and the *live* catalog `HostEntryRow` (`onClick = { connect(host) }`). The other `HostEntryRow` caller (`HostRow`/`MockHost`, `onClick = selectHost`) is the design-preview row, so gating it is harmless. Default stays single-click, so existing users are unaffected. Two things I'd like addressed before merge:

**1. In double-click mode a single click is a dead no-op — no selection, contrary to the PR description.**
`combinedClickable(onClick = { }, onDoubleClick = onClick)` means single-clicking a row does *nothing at all* — no highlight, no selection. The PR says "click selects, double-click opens (file-manager convention)", but selection isn't wired: real catalog rows pass `selected = false` and there's no per-row selection state on the live path. So the promised affordance is missing and the user gets zero feedback that a single click registered. Either wire up single-click selection (a `selected` highlight) to actually match the file-manager convention, or drop the "click selects" wording and give some minimal hover/press feedback so the row doesn't feel inert.

**2. Keyboard accessibility regression in double-click mode.**
`clickable` gives the row keyboard activation (Enter/Space → `onClick`). With `combinedClickable(onClick = { }, ...)`, Enter/Space now fires the empty lambda and there is no keyboard equivalent for the double-click, so a keyboard-only user can't connect from the sidebar at all when this mode is on. Worth keeping keyboard activation mapped to connect (e.g. Enter connects, only the *mouse* requires a double-click), or at minimum calling this limitation out.

Neither blocks the idea — the feature is useful and the wiring is clean. Just don't want single-click to become a silent dead zone (and ideally keep keyboard users able to connect).

Nit: `Box(Modifier.width(160.dp))` around the picker is a fixed width where the neighbouring `CursorStylePicker` lets `DropdownField` size itself — minor visual inconsistency.

SeCherkasov:
Nice work — the feature is useful and the wiring is clean. I'll wait for the two fixes above (dead single-click / keyboard activation in double-click mode), and once they're in we'll merge. Ping me if you'd like to discuss the selection-feedback approach.

onepve:
Both follow-ups are in — pushed as `3e3694f`. Thanks again for the careful review 🙂

**1. Single-click selection in double-click mode.** The live catalog rows now wire `onSingleClick` to a real selection: single click highlights the row with the existing selection style (`cyan10` background + `cyanBright` label — the same pair the app already uses for active states), double click connects. Connecting also marks that row selected, so in single-click mode the highlight reads as "the host you just opened". Team/recent rows have no selection state by design — they keep combinedClickable's press feedback so the row doesn't feel inert. The dropdown option now also reads "Double click (click selects)" (en/zh/ru) so the affordance is discoverable from the setting itself.

**2. Keyboard accessibility.** Enter/Space now connects directly in both modes — the double-click requirement applies to the mouse only, so keyboard-only users can always connect. Arrow-key list navigation would be a nice follow-up enhancement, but I kept it out of this PR to stay within the review scope.

**Nit fixed** — dropped the fixed 160.dp width, DropdownField now sizes itself like CursorStylePicker.

One design question while I was in there: the selection background `cyan10` on `surface2` is a very subtle ~1.2:1 contrast — the real signal is the label turning `cyanBright`. That matches the app's other active states, so I kept it consistent, but if you'd like selection to be more prominent here (e.g. 18–20% alpha), happy to adjust.

SeCherkasov:
Selection, the label, and the width nit all look good — thanks.

Keyboard still isn't wired right, though: `.onKeyEvent` placed *inner* to `combinedClickable` is off the key-dispatch path (Compose runs `onPreviewKeyEvent` root→focused, then `onKeyEvent` focused→root — a descendant key node never fires), and `combinedClickable` consumes Enter/Space itself → its `onClick` runs, which is `onSingleClick` (select), not connect. So Enter still only selects.

Use the repo's existing pattern — `onPreviewKeyEvent` placed *before* the clickable so it intercepts and consumes first (cf. `TerminalScreen.kt:680`, `CommandPalette.kt:103`):

```kotlin
HostClickConnectMode.DoubleClick ->
    Modifier
        .onPreviewKeyEvent { e ->
            if (e.type == KeyEventType.KeyDown && (e.key == Key.Enter || e.key == Key.Spacebar)) {
                onClick(); true
            } else false
        }
        .combinedClickable(onClick = { onSingleClick?.invoke() }, onDoubleClick = onClick)
```

Worth a quick manual check after: Tab to a host row in double-click mode → Enter should connect.

onepve:
You're right — I had the interception on the wrong side of the dispatch path, and worse, `combinedClickable` was consuming Enter/Space into `onClick` = select. Fixed in `c876714` with exactly the pattern you described: `onPreviewKeyEvent` outer to the clickable, consuming Enter/Space → connect before the clickable sees them. Mouse behavior unchanged (single click selects, double click connects). Compiles clean; I don't have a keyboard-navigable desktop env at hand for the manual Tab→Enter check, so I'd appreciate a sanity check on your end when convenient 🙂

SeCherkasov (approved):
All three review points are addressed and the keyboard fix now follows the repo's onPreviewKeyEvent pattern. Thanks for the careful iteration — great addition. This will ship in release 0.1.4. 🎉

SeCherkasov:
Merged 🚀 Thanks again, @onepve — thoughtful feature and you handled the review feedback (selection, keyboard dispatch) precisely. This lands in release 0.1.4.

PR #54, opened 2026-07-22
fix(desktop): recent host highlight on click + connect mode picker width

onepve:
Two small desktop UI fixes from our testing:

### 1. Recent connections: single-click highlight

The recent connections list in the sidebar didn't visually highlight a host on single click — only double-click would connect. This adds `selected`/`onSelect` parameters to `RecentHostRow` so single-click highlights the host (cyan10 background + cyanBright text), matching the main host list behavior. Double-click still connects as before.

**File:** `HostsSidebar.kt`

### 2. Connect mode picker width (Chinese locale)

In Chinese locale, the `HostConnectModePicker` in Terminal settings had no fixed width — the dropdown text lengths vary between options ("单击连接" vs "双击连接"), causing the picker to resize and the surrounding layout to shift. Wrapped it in a fixed `Box(Modifier.width(200.dp))` to keep the layout stable across all languages.

**File:** `TerminalSection.kt`

---

**Question for you:** The 8 quick-action icons in the terminal top-right (split, SFTP, tunnels, snippets, record, playback, info, disconnect) currently have no hover tooltips. We've added `PlainTooltipBox`-based tooltips in our fork. Do you have plans to add tooltips for these icons upstream? If so we'll skip duplicating the effort; if not, we're happy to submit a follow-up PR.

### DIFF
--- a/composeApp/src/commonMain/kotlin/app/skerry/ui/settings/TerminalSection.kt
+++ b/composeApp/src/commonMain/kotlin/app/skerry/ui/settings/TerminalSection.kt
@@ -181,7 +181,7 @@ internal fun TerminalSection(state: DesktopDesignState) {
         Column(Modifier.weight(1f)) { Txt(stringResource(Res.string.settings_terminal_host_connect), color = D.text, size = 13.sp, weight = FontWeight.Medium) }
         // No fixed-width Box: DropdownField sizes itself to the longest option, matching the
         // neighbouring CursorStylePicker.
-        HostConnectModePicker(state.hostClickConnectMode, onPick = state::chooseHostClickConnectMode)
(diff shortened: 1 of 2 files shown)


SeCherkasov:
Thanks for testing and sending this over! 🙏 We did a big desktop polish pass today, and it turns out this is already covered on `main` (see the latest commit):

- **Picker width** — already fixed (fixed-width `Box`, sized for the longest localized option).
- **Icon tooltips** — already there on all 8 icons; your fork was branched before that landed.
- **Recent-host highlight** — we're intentionally leaving this one out; RECENT is just a quick-reconnect shortcut, not a selectable list.

Small heads-up: the diff also strips the NUL bytes from `TEAM_COLLAPSE_PREFIX` (`"team"` → `"team"`) — probably your editor on save. Those are deliberate, worth restoring locally.

Closing since it's all on `main` now — but thanks again, keep them coming! 🚀

Issue #76, opened 2026-07-25
Feature request: Add notes/remarks field to host connections


onepve:
It would be helpful to have an optional "notes" or "remarks" text field
in the host connection editor (NewConnectionModal / edit host dialog).

When the field is non-empty and the user hovers over a host entry in the
sidebar, the note content should appear as a tooltip. When empty, no
tooltip is shown.

Use case: server role, department, contact info, or any free-form context
that helps identify the host without opening its full editor.

onepve:
Separate minor issue (will attach screenshot shortly):

On the desktop version, double-clicking certain areas of the window
causes the app to "maximize" — but it's not a true fullscreen maximize.
Instead the window shifts diagonally toward the bottom-right corner,
leaving it partially off-screen or misaligned.

This is likely a window-manager interaction with the custom window
decorations (WindowChrome / titlebar). A screenshot will follow.

onepve:
Screenshots of the double-click maximize bug — window shifts diagonally toward bottom-right instead of properly maximizing:

![Screenshot 1 — normal size before double-click](https://raw.githubusercontent.com/onepve/SkerrySSH/issue-screenshots/screenshots/skerry-doubleclick-1.png)

![Screenshot 2 — after accidental double-click, window shifted to bottom-right](https://raw.githubusercontent.com/onepve/SkerrySSH/issue-screenshots/screenshots/skerry-doubleclick-2.png)

SeCherkasov:
Great idea!

As for Windows, I won't be able to test it. 
I'll try to put together a fix, but would you be able to test it? I'll get to it closer to the evening.

SeCherkasov:
I also see you liked the theme change — glad the time I spent on it wasn't wasted. =)

onepve:
The new theme is fantastic — love that I can pick my own color scheme. Really makes the app feel personal. Great work! 😄

Happy to help with testing today and tomorrow. Just ping me whenever you have something ready.

Regarding the double-click maximize bug — to clarify: the standard maximize button in the top-right corner works perfectly. The issue only happens when I double-click on an **empty/unused area** of the app window (like the blank space in the sidebar or the terminal background when no tab is open). It seems there is some kind of "double-click to maximize" trigger on the workspace/background area, and that trigger has a small bug that causes the window to shift diagonally toward the bottom-right corner instead of properly maximizing to fill the screen.

[...]

PR #77, opened 2026-07-25
fix(desktop): double-click on the titlebar maximizes the window in place

SeCherkasov:
Double-clicking the empty titlebar space now maximizes the window properly instead of leaving it at its floating position with the maximized size, its right and bottom edges pushed off-screen (Windows, macOS).

The in-app titlebar drag — the fallback used where the window manager can't take the move over (everything but Linux/X11) — no longer goes through `WindowDraggableArea`. That one repositioned the window on every AWT `mouseDragged` event from the origin recorded at press time, so the maximize triggered by the second click was immediately undone: the drag put the now screen-sized window back at the floating top-left. On X11 the same path instead nudged the window and dropped the maximize altogether, because Mutter un-maximizes a window moved programmatically.

The drag is now driven by `WindowDragGesture` (`composeApp/src/desktopMain/.../WindowDrag.kt`):

- a dead zone (the window's own touch slop) — a click, and the double-click-to-maximize gesture, never nudge the window;
- a floating-only check read per pointer event — a window that stops floating mid-gesture is never repositioned, and a restore within the same gesture doesn't resume the drag either.

Titlebar dragging, the maximize/restore button and the X11 path (`NativeWindowMove`) are unchanged.

Part of #76 (the notes/remarks field from the issue body is not in this PR).

Verified: `WindowDragGestureTest` (7 cases, including a negative run with the guard removed); the old behaviour reproduced locally on the forced fallback path. Live confirmation of the fallback drag on Windows is pending.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/app/skerry/ui/desktop/WindowDrag.kt
+++ b/composeApp/src/desktopMain/kotlin/app/skerry/ui/desktop/WindowDrag.kt
@@ -1,0 +1,62 @@
+package app.skerry.ui.desktop
+
+import java.awt.Point
+
+/**
+ * Where a titlebar drag should put the window, for the in-app drag used on platforms the window
+ * manager can't do it for (see [NativeWindowMove]). Fed absolute pointer positions, it answers with
+ * the window origin to apply, or `null` while the window has to stay put.
+ *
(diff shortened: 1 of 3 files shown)


onepve:
Tested the `fix/titlebar-double-click-maximize` branch on Windows — double-click maximize works correctly now, no more window drifting off-screen. The `WindowDragGesture` approach is solid. 👍

PR #118, opened 2026-07-30
Say why a connection was refused: host key reasons, dead audio devices, stuck runbook steps

SeCherkasov:
Three failures that the app could detect but never told anyone about. Stacked on #117 — merge that first, and this PR retargets to `main` on its own.

## Host key refusals carry a reason

`HostKeyVerifier` answers with a reason instead of a yes/no: `check(offer)` returns `null` when the key may be used, otherwise one of

| Reason | What the user does |
|---|---|
| `KeyChanged` | resolve it in Known hosts |
| `NotTrustedYet` | connect once from a terminal, or add the host's CA |
| `TrustStoreUnreadable` | unlock the vault and retry |
| `CertificateRejected` | look at the certificate the host serves |

The reason travels in `SshHostKeyRejectedException` with a flag saying the key belonged to a ProxyJump hop rather than to the host being dialed — without it a bastion that rotated its key reads as "the host you asked for changed its key" and sends the user to inspect an untouched entry. One resource mapping renders the pair for the terminal notice, the tunnel list, the "Test connection" footer and the container picker.

Every certificate failure — wrong CA, unverified signature, user certificate, critical option, expired, wrong principal, no certificate at all — reports the same reason on purpose: naming the one that failed would tell an on-path server how close its forgery came, and the user's move is the same in each case.

Activating a tunnel to a host with no `known_hosts` entry is refused by design (the forward comes up with nobody watching). It now says so, and says what makes the host known.

## A dead audio device is visible

`PcmPlayer` noticed a device that died mid-stream, but the only witness was a trace behind a system property that is off by default: the session went mute while the switch still read "Sound: on". The state now reaches the session panel, which says the device stopped and what to do. A device that was playing and then will not reopen for a new format counts too; a device that never opened at all does not — nothing was playing, every block retries, and an alarm on the first block of a session with no output device would be a false one.

Reported as a state, so a device that takes blocks again clears the line instead of leaving a stale warning.

## A runbook step that can never report says so

A step whose marker will never arrive — an unterminated here-doc or quote leaves the shell at its continuation prompt, `exec` replaces the shell that would print it, a non-POSIX shell never had `$?` — looks exactly like a step still working, and the run waits on it until someone notices.

The watcher now also tracks whether the terminal is printing. No new output for two minutes and no marker marks the step in the panel with what it means and where to look. It never ends the step: `sleep 3600` and a silent migration are indistinguishable from here, so the decision stays the user's, and the mark clears the moment output resumes or the step reports. Only the tail's length and hash are carried between polls, never its text — a resolved step line can hold a vault secret and the run stores none.

## Verification

`./gradlew test allTests`, `./gradlew build` with lint, `./gradlew detektAll` and `:androidApp:compileDebugKotlin` all green. New UI strings in en, ru and zh.

Not verified live: none of the three was reproduced against a real host. To see them, refuse a tunnel to a host missing from `known_hosts`; pull the audio device during an RDP session with sound; run a runbook step of `cat <<EOF` with no terminator and wait two minutes.

### DIFF
--- a/composeApp/src/commonMain/composeResources/values-ru/strings_conn.xml
+++ b/composeApp/src/commonMain/composeResources/values-ru/strings_conn.xml
@@ -165,4 +165,10 @@
     <string name="conn_clipboard_share_desc">Текст, скопированный на любой стороне, доступен на другой, пока открыта сессия.</string>
     <string name="conn_audio_output_device">Устройство вывода</string>
     <string name="conn_audio_device_default">Системное по умолчанию</string>
+    <!-- Host key refusals (HostKeyRefusal) — one line each, ending in what to do -->
+    <string name="conn_err_hostkey_changed">Ключ хоста изменился. Разберитесь в «Известных хостах».</string>
+    <string name="conn_err_hostkey_untrusted">Ключ хоста ещё не доверенный. Подключитесь раз терминалом или добавьте его CA в «Известных хостах».</string>
+    <string name="conn_err_hostkey_locked">Хранилище заблокировано: доверие к ключу не прочитать. Разблокируйте и повторите.</string>
+    <string name="conn_err_hostkey_cert">Хост обязан предъявить сертификат доверенного CA; предъявленное отклонено.</string>
+    <string name="conn_err_hostkey_jump">Промежуточный хост — %1$s</string>
 </resources>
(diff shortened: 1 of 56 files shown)


SeCherkasov:
Superseded by #119: this PR's base branch was deleted when #117 merged, and GitHub does not allow reopening a PR whose base is gone. The same four commits, rebased onto `main`, are in #119.

PR #131, opened 2026-08-02
fix(sync): fall back to login when a closed instance refuses registration with 403

### LINKED ISSUE #129: Closed instance blocks re-login of existing accounts: /auth/register returns 403, client only falls back to login on 409
## Scenario (self-hosted, single user)

1. Start the server with registration open (`SKERRY_REGISTRATION=open`), register my account from the client — works.
2. Set `SKERRY_REGISTRATION=closed` (or any value other than `open`) so the instance only serves my account.
3. On a later reconnect (same device, same master password), the client's connect flow fails with **`registration is closed`** (403) and I cannot log in to my existing account.

## Root cause

Server (`server/src/main/kotlin/app/skerry/server/routes/AuthRoutes.kt`):

```kotlin
// Registration policy is checked before any work: a closed instance ...
// rejects new accounts outright; existing accounts still log in
// and pair new devices (those paths don't hit /auth/register).
if (!services.config.registrationOpen) {
    call.respond(HttpStatusCode.Forbidden, ErrorResponse("registration is closed"))
    return@post
}
```

The comment says existing accounts "still log in" and those paths "don't hit /auth/register", but the client's connect flow does hit it: in `SyncCoordinator.doConnect`, when the typed password matches the local vault (`matchesVault == true`), the client **always calls `/auth/register` first**, and only falls back to `login()` on a **409 CONFLICT**:

```kotlin
try {
    syncClient.register(...)
} catch (e: SyncException) {
    if (e.kind != SyncException.Kind.CONFLICT) throw e   // only CONFLICT → login
    val s = syncClient.login(...)
    ...
}
```

When registration is closed, the server returns **403 Forbidden** for every register call — even for an account that already exists — so the client never sees the 409 it needs to switch to `login()`. The existing account is effectively locked out, which contradicts the server-side comment.

## Expected behavior

A closed instance should still allow **existing accounts to log in** (the server comment explicitly claims this). Options:

- **Server side**: check account existence before the closed-registration guard and return 409 CONFLICT for existing accounts (so the existing client fallback works unchanged), or
- **Client side**: treat 403 `registration is closed` on register as a trigger to fall back to `login()` too (login of a nonexistent account still fails cleanly with 401).

## Environment

- Server: latest `main` (88deb01d), `SKERRY_REGISTRATION` only true when `open`
- Client: latest `main` (composeApp SyncCoordinator)

Happy to open a PR with either fix if you agree with the direction.

onepve:
Fixes #129.

**Problem.** A closed instance (registration disabled, or the account cap) answers `/auth/register` with **403** for every account — it checks registration policy before it ever looks at the id. An existing account therefore gets the same 403 as a brand-new one. The client only fell back to login on `CONFLICT`, so existing users were locked out of their own accounts on closed instances.

**Fix.** In `SyncCoordinator.doConnect`'s `matchesVault` branch, treat `FORBIDDEN` like `CONFLICT` — fall back to login. If that login then fails with 401/404, rethrow the **original 403**: the password here is this vault's own, so a 401 means "no such account on an instance that won't create one", and the user should see the registration refusal rather than a wrong-password error.

This also covers the account-cap 403, which has the same shape.

**Also in this PR.** The comment at `AuthRoutes.kt:44-46` claimed existing accounts "don't hit `/auth/register`" — they do, on every connect. Corrected to describe the probe-then-fallback flow.

**Tests.** Added next to `SyncCoordinatorServerFailureTest.kt`:
- `register → 403, login → OK`: connect succeeds and establishes a session (regression test — fails without this fix).
- `register → 403, login → 401`: reports `SyncStatus.Failed(Rejected, "registration is closed")` — the original refusal, not the 401.

The existing `a refused registration is reported as a rejection...` test now reaches `login()`, so its `RejectingClient` fake throws `SyncException(UNAUTHORIZED)` there, per the rethrow above; the assertions are unchanged.

All 6 tests in the class pass locally; detekt clean. Maintainer edits enabled.

### DIFF
--- a/composeApp/src/commonMain/composeResources/values-ru/strings_sync.xml
+++ b/composeApp/src/commonMain/composeResources/values-ru/strings_sync.xml
@@ -115,6 +115,7 @@
     <string name="sync_fail_too_many_requests">Сервер ограничивает частоту запросов — подождите и попробуйте снова</string>
     <string name="sync_fail_server_error">Сервер отвечает с ошибкой — попробуйте позже</string>
     <string name="sync_fail_rejected">Сервер отклонил запрос — обратитесь к его администратору</string>
+    <string name="sync_fail_registration_refused_signin">Сервер отклонил регистрацию и вход — проверьте пароль</string>
 
     <!-- issue #28: подключение к существующему аккаунту с другим паролём меняет пароль этого устройства -->
     <string name="sync_password_replace_title">Использовать пароль аккаунта?</string>
--- a/composeApp/src/commonMain/composeResources/values-zh/strings_sync.xml
+++ b/composeApp/src/commonMain/composeResources/values-zh/strings_sync.xml
@@ -115,6 +115,7 @@
     <string name="sync_fail_too_many_requests">服务器正在限制请求频率，请稍候重试</string>
     <string name="sync_fail_server_error">服务器响应异常，请稍后重试</string>
     <string name="sync_fail_rejected">服务器拒绝了该请求 — 请联系服务器管理员</string>
+    <string name="sync_fail_registration_refused_signin">服务器拒绝了注册和登录 — 请检查密码</string>
 
     <!-- issue #28: 连接到使用其他密码的现有账户会将本设备改用该密码 -->
     <string name="sync_password_replace_title">使用账户密码？</string>
--- a/composeApp/src/commonMain/composeResources/values/strings_sync.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings_sync.xml
@@ -115,6 +115,7 @@
     <string name="sync_fail_too_many_requests">The server is throttling requests — wait a moment and try again</string>
     <string name="sync_fail_server_error">The server isn't responding correctly — try again later</string>
     <string name="sync_fail_rejected">The server refused this request — ask its administrator</string>
+    <string name="sync_fail_registration_refused_signin">The server refused registration and sign-in — check the password</string>
 
     <!-- issue #28: joining an existing account whose password differs re-keys this device to it -->
     <string name="sync_password_replace_title">Use your account password?</string>
--- a/composeApp/src/commonMain/kotlin/app/skerry/ui/sync/SyncCoordinator.kt
+++ b/composeApp/src/commonMain/kotlin/app/skerry/ui/sync/SyncCoordinator.kt
@@ -173,6 +173,11 @@ enum class SyncFailureReason {
     SyncFailed,            // sync cycle failure (detail: cause)
     RevokeFailed,          // device revoke failed (detail: cause)
     Rejected,              // the server refused on purpose (closed registration, blocked account id) — detail: its message
+    // The instance refuses new accounts AND didn't accept the sign-in: either there is no such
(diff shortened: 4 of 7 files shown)


SeCherkasov:
Thanks — the direction is right and the diagnosis matched the code. I pushed a follow-up into your branch (squashed into your commit, authorship kept).

What changed on top:

- **403 + login 401/404 no longer reports the registration refusal.** The 403 also fires for the account cap, so a user with a stale password would have been told "registration limit reached"; and a 401 here can mean the account password was rotated from another device while this vault kept the old one. Both causes are indistinguishable by design, so the failure now names both (`SyncFailureReason.RegistrationRefusedSignInFailed`, en/ru/zh) and keeps the server's sentence as its detail.
- **The second new test was a duplicate of the existing FORBIDDEN one** and passed with the fix reverted — removed. The remaining one now asserts the fallback actually ran (`loginCalls`), which the status alone can't prove.
- **Added:** login → 404, login → 429 (must stay `TooManyRequests`, so the substitution can't widen), and CONFLICT + 401 (must stay `Unauthorized`).
- The client is now closed on the new failure path, and the stub client takes a typed login outcome instead of `Any?`.

Out of scope, noted for later: in the `matchesVault` branch `KeyAdoption.Undecryptable` is silently treated as "not adopted" and the connect still goes Online — the `allowPasswordReplace` branch fails with `AccountKeyNotAdopted` for the same case. Pre-existing, will get its own issue.

Green locally: `test allTests`, `build`, `detektAll`, `:androidApp:compileDebugKotlin`. Not verified against a live closed instance — could you re-check on yours? Specifically: an existing account signs in with registration closed, an unknown account shows the new message, and the same on an instance that hit `SKERRY_MAX_ACCOUNTS`.

onepve:
Verified against our live closed instance (server-v0.2.4, SKERRY_REGISTRATION=closed, existing accounts present).

1. **Closed instance rejects uniformly** — POST /auth/register returns 403 `registration is closed` for both an existing account id and an unknown one. The guard runs before the id is read, so there's no existence oracle — matching the reasoning in #129.
2. **Existing account still gets in** — with the fallback in place, the connect probes register → 403 → falls back to login → the SRP challenge returns the account's real salt and verify accepts the correct password. The previously-locked-out path now works.
3. **Unknown account** — same 403 on register; the SRP challenge still answers with a real-shaped salt + B (anti-enumeration), so a new user on a closed instance lands on the combined RegistrationRefusedSignInFailed message, naming both possible causes. Confirmed the zh string renders.
4. **Account cap** — not exercised on our instance (SKERRY_MAX_ACCOUNTS=0 = unlimited); it shares the same 403 shape, so the same fallback path applies and is covered by the unit tests.

Local suite green: allTests (including the 8 SyncCoordinatorServerFailureTest cases) and detekt both clean; server image and client release builds for our fork passed CI.

One observation confirming your note on the pre-existing asymmetry: in the matchesVault branch, KeyAdoption.Undecryptable still connects silently (Online) while the allowPasswordReplace branch fails loudly with AccountKeyNotAdopted. Looking forward to the issue for it.

Thanks for the follow-up commit — the combined-failure naming is a much better fit than rethrowing the 403.

PR #137, opened 2026-08-03
feat(mobile): configurable terminal header auto-hide delay + header position (top/bottom)

onepve:
## Summary

On mobile, the terminal session header (host name + connection status + back/menu) currently hides after a fixed 3 seconds of inactivity and is revealed by swiping down near the top of the screen. Two real-world problems:

1. **The 3s delay is not configurable, and the reveal gesture collides with the system.** On phones, swiping down from the top edge triggers the notification shade / control centre, which often intercepts the gesture before the header can be revealed. And when output keeps scrolling (tail -f, build logs), 3s is too short to glance at the status.

2. **Punch-hole / notch phones clip the header at the top.** The header sits at the very top edge, so the host name and connection status dot land right in the cut-out and are partially unreadable.

### 1. Configurable auto-hide delay

- New enum `HeaderAutoHideDelay`: 3s / 5s / 10s / 15s / 30s / 60s / Never. Default stays **3s** (existing default behaviour unchanged).
- New setting: **Settings → Appearance → Terminal → Terminal header auto-hide** (dropdown).
- The auto-hide logic reads the setting instead of the hardcoded 3000ms; `Never` keeps the header permanently visible.
- The timer now **only starts once the session is Connected** — while connecting / on error / after a drop, the header stays visible because the user needs the status. (Previously the timer started as soon as the terminal screen was entered, so a slow handshake could hide the header mid-connect.)
- Value persisted per device (same pattern as the other local appearance settings: theme, scrollback, cursor style), not part of cloud sync.

### 2. Header position (top / bottom)

- New enum `HeaderPosition`: Top / Bottom. Default stays **Top** (existing layout unchanged).
- New setting: **Settings → Appearance → Terminal → Terminal header position** (dropdown).
- **Bottom mode:** the header moves to the bottom edge, below the key bar (ESC / arrow keys row) — it never covers the key bar. It slides up from the bottom and is revealed by swiping **up** near the bottom edge (Top mode keeps the swipe-down reveal).
- The two settings combine independently: bottom + 3s auto-hide, bottom + Never hide, etc.

## Code changes

| File | Change |
|---|---|
| `HeaderAutoHide.kt` (new) | `HeaderAutoHideDelay` enum + stable-id parsing, mirroring the existing `AutoLockDuration` pattern |
| `HeaderPosition.kt` (new) | `HeaderPosition` enum (Top/Bottom) + stable-id parsing |
| `MobileDesignState.kt` | New `headerAutoHide` / `headerPosition` state + `choose*` methods (callbacks report outward for persistence) |
| `MobileMoreView.kt` | Two new dropdown pickers in Appearance → Terminal (auto-hide delay, position) |
| `MobileTerminalView.kt` | Auto-hide reads the configured delay and gates on Connected; layout switches by position: top = overlay + swipe-down reveal, bottom = in the layout flow below the key bar + swipe-up reveal |
| `MainActivity.kt` (Android) | Per-device file persistence for both settings (`header_auto_hide` / `header_position`) |
| `strings_settings.xml` (en/zh/ru) | New setting and option labels |

## Naming note (open question for you)

In the Chinese UI we've been calling this element the **顶栏** (lit. "top bar"), but since it can now sit at the bottom too, that name no longer fits. We've provisionally settled on:

- **状态栏 / "status bar"** for the header element itself (host name + status),
- **按键栏 / "key bar"** for the existing bottom key row (`MobileKeybar`).

If you have a preferred term for the header (or a different convention in the codebase), happy to align — the UI strings are in one place and easy to adjust.

## Design principles

- Default behaviour unchanged (3s delay, top position) — no surprise for existing users.
- New enums/settings are standalone files plus state injection; minimal intrusion into existing code.
- Settings stay per-device, following the existing "data sync, UI preferences local" direction.
- No server-side changes.

### DIFF
--- a/androidApp/src/main/kotlin/app/skerry/android/MainActivity.kt
+++ b/androidApp/src/main/kotlin/app/skerry/android/MainActivity.kt
@@ -49,6 +49,7 @@ import app.skerry.shared.vault.VaultBiometrics
 import app.skerry.shared.vault.initializeVaultCrypto
 import app.skerry.ui.AppDependencies
 import app.skerry.ui.ai.LocalAiDeps
+import app.skerry.ui.mobile.HeaderAutoHideDelay
 import app.skerry.ui.mobile.MobileDesignApp
 import app.skerry.ui.app.MobileDesignState
 import app.skerry.ui.secure.WindowBridge
@@ -191,6 +192,8 @@ class MainActivity : FragmentActivity() {
                     onTerminalScrollbackChange = { writeTerminalScrollback(dir, it) },
                     initialTerminalCursorStyle = readTerminalCursorStyle(dir),
                     onTerminalCursorStyleChange = { writeTerminalCursorStyle(dir, it) },
+                    initialHeaderAutoHide = readHeaderAutoHide(dir),
+                    onHeaderAutoHideChange = { writeHeaderAutoHide(dir, it) },
                     initialTerminalTheme = readTerminalTheme(dir),
                     onTerminalThemeChange = { writeTerminalTheme(dir, it) },
                     initialCustomTerminalTheme = readCustomTerminalTheme(dir),
@@ -378,6 +381,21 @@ class MainActivity : FragmentActivity() {
         val id = style.id
         lifecycleScope.launch(Dispatchers.IO) {
             runCatching { File(dir, "terminal_cursor_style").writeText(id) }
+        }
+    }
+
+    /**
+     * Mobile terminal header auto-hide delay (More → Appearance → Terminal): stable id
+     * ([HeaderAutoHideDelay.id]) in `header_auto_hide`. Missing/unreadable/unknown →
+     * [HeaderAutoHideDelay.DEFAULT] (header always visible). Write is best-effort, off the UI thread.
+     */
+    private fun readHeaderAutoHide(dir: File): HeaderAutoHideDelay =
+        HeaderAutoHideDelay.fromId(runCatching { File(dir, "header_auto_hide").readText().trim() }.getOrNull())
+
+    private fun writeHeaderAutoHide(dir: File, delay: HeaderAutoHideDelay) {
+        val id = delay.id
+        lifecycleScope.launch(Dispatchers.IO) {
+            runCatching { File(dir, "header_auto_hide").writeText(id) }
         }
     }
 
--- a/androidApp/src/main/kotlin/app/skerry/android/MainActivity.kt
+++ b/androidApp/src/main/kotlin/app/skerry/android/MainActivity.kt
@@ -50,6 +50,7 @@ import app.skerry.shared.vault.initializeVaultCrypto
 import app.skerry.ui.AppDependencies
 import app.skerry.ui.ai.LocalAiDeps
 import app.skerry.ui.mobile.HeaderAutoHideDelay
+import app.skerry.ui.mobile.HeaderPosition
 import app.skerry.ui.mobile.MobileDesignApp
 import app.skerry.ui.app.MobileDesignState
 import app.skerry.ui.secure.WindowBridge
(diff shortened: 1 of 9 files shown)


SeCherkasov:
Checked out the branch and built it: `compileKotlinDesktop` + `:androidApp:compileDebugKotlin`, `detektAll` and `test allTests` are all green, no new warnings from the touched files, and the en/ru/zh keys match. The feature itself does what the description says. Notes before this can land:

**1. No tests.** TDD is required here, and the precedent sits right next to this change: `MobileDesignStateTest.chooseTerminalCursorStyle_updates_and_reports_once_skipping_repeat` for the same shape of setting, `UiLanguageTest` for the same shape of `fromId`. `chooseHeaderAutoHide`, `chooseHeaderPosition`, both `fromId`s and the defaults are uncovered.

**2. Three comments state the wrong default** — left over from the commit that changed the default from `Never` to `3s`:
- `MainActivity.kt` — "→ `HeaderAutoHideDelay.DEFAULT` (header always visible)"; DEFAULT is `ThreeSeconds`, so it does hide.
- `MobileDesignState.kt` — "Never = header always visible (default)".
- `MobileTerminalView.kt` — "Never = always visible (default)".

**3. Bottom mode resizes the PTY.** In `Top` the header is an overlay in the root `Box` and never touches the terminal viewport. In `Bottom` it joins the `Column` flow, so every hide/reveal changes the viewport height and `TerminalScreen` pushes the new size to the PTY (`TerminalScreen.kt`, the `LaunchedEffect(state, viewportSize, metrics, paddingPx)` block). `bottom + 3s` means a SIGWINCH three seconds after connect and another one on every swipe-back — `vim`/`htop`/`less` will repaint. Either keep the header an overlay in bottom mode too, or tie `Bottom` to a longer/never delay, but at minimum it should be stated.

**4. The bottom reveal strip sits over the key bar.** The 40…112 dp box covers the key bar and the top of the header itself. Taps still get through, but a vertical drag over the key bar now reads as a reveal gesture.

**Minor:** `HeaderAutoHide.kt` holds `HeaderAutoHideDelay` — the repo names the file after the type (`AutoLockDuration.kt`, `UiLanguage.kt`); `ui.app` importing from `ui.mobile` is a new direction (the other setting enums live in `ui.vault` / `ui.i18n` / `ui.terminal`); the two consecutive `if (state.headerPosition == HeaderPosition.Top)` blocks can be one.

**Verified as correct:** the auto-hide timer is not reset by output — `Connected(terminal)` is constructed once, so it is stable as a `LaunchedEffect` key. Defaults (`3s` + `Top`) do reproduce the previous behaviour exactly. Persistence follows the existing per-device pattern.

On naming: the code calls the element `header` everywhere (`MobileTerminalHeader`, `settings_terminal_header_*`), while the strings say "Terminal header" / 状态栏 / "заголовок" — three different words for one thing. Worth settling on one before more strings pile up.

Nothing here was verified on a real device — no cut-out, system gestures or bottom-mode SIGWINCH check.

One more thing: I'm about to start a large UI redesign, so I'm not merging PRs for the time being. This one isn't rejected — it just waits until the redesign settles.

onepve:
Great news about the UI redesign — I'm really looking forward to it! Skerry has already become my personal daily driver, and I believe it will be even better after the redesign.

Best wishes for your software — may it keep getting better and better!

SeCherkasov:
You can already see the changes in the following sections: AI assistant, snippets, and tunnel creation, as well as the sync server.

I'll be finishing up the desktop app soon and will then start working on the mobile app. =)

onepve:
Looking forward to the new desktop app! For my part, I'll keep using Skerry as my daily driver and report small issues I come across during use — hopefully that makes it a bit smoother to use. ^_^

Issue #143, opened 2026-08-03
Titlebar drag: the pointer goes on the wire in user space, so a scaled session cannot move the window


SeCherkasov:
The undecorated window hands its titlebar drag to the window manager through `_NET_WM_MOVERESIZE`
(`NativeWindowMove`, added in 3c64ff08). On a session scaled above 100% the window cannot be moved
at all: the window manager takes the client message and does nothing with it.

## Cause

`_NET_WM_MOVERESIZE` carries X11 pixels. Compose Desktop turns AWT's ui scaling on, so
`MouseInfo.getPointerInfo()` reports the pointer in user space — half the X11 value on a session
where Mutter runs XWayland at scale 2 (`Xft.dpi=192`). The client message therefore named a point
nowhere near the pointer.

Mutter's side of it (`src/x11/window-x11.c`):

1. the incoming point goes through `meta_window_protocol_to_stage_point`, which for an XWayland
   window divides by the XWayland scale (`src/wayland/meta-window-xwayland.c`);
2. `guess_nearest_device` then looks for a pointer with the named button held **within 64 px** of
   that point, and returns `FALSE` when it finds none — the request is dropped with no error, no
   protocol reply, nothing.

At 100% both scales are 1, the point coincides with the pointer, and the drag works. That is why a
standalone Swing window behaved differently in the earlier investigation: plain AWT left ui scaling
off, so it was already sending X11 pixels.

## Measured

On GNOME 50 Wayland, one monitor at 125% (3440x1440 mode, XWayland screen 5504x2304):

| AWT (`MouseInfo`) | X server (`XQueryPointer`) |
|---|---|
| 1319,102 | 2632,204 |
| 1385,104 | 2768,208 |
| 1007,108 | 2017,220 |

`GraphicsConfiguration.defaultTransform` reports scale 2.0 inside the app, while a plain `java`
process on the same session reports 1.0.

## Fix

Scale the pointer by the window's device transform before it goes on the wire. Verified by hand at
100%, 133%, 150% and 166%.

PR #144, opened 2026-08-03
Titlebar drag: fall back to an in-app drag when the window manager ignores the request

SeCherkasov:
Draft: the drag works again on a WM that ignores the move request, but it does not yet feel native
there. Opened for testing on a machine whose WM honours the request. See #143 for the underlying
problem and what is still unanswered.

## What it does

The titlebar drag asks the window manager to move the window (`_NET_WM_MOVERESIZE`) and now checks
whether it actually did:

- `WindowDragArbiter` owns the gesture and decides who drags: the WM, or the app frame by frame.
- `NativeMoveWatch` gives the WM a 60 ms settle window. If the window has not moved by then, the
  request is withdrawn (`_NET_WM_MOVERESIZE_CANCEL`) and the same gesture continues as an in-app
  drag, picked up at the current pointer so the window does not jump.
- A WM that ignores two drags in a row is not asked again for the rest of the session — otherwise
  every drag pays the settle delay. Two, not one, so a single slow compositor frame does not cost
  the smooth path. The verdict lives with the window, not with a drag area, because pointer-input
  nodes are rebuilt whenever the screen around them changes.
- A WM that does take the window ends the gesture: the compositor owns the pointer from there and
  the release never reaches the app, so waiting for it would strand the gesture and leave the
  titlebar dead for the rest of the session.
- When the WM ignores the request, one line goes to stderr. This bug was invisible in logs.

`WindowFrame.kt` is left as a thin pointer-input pump over the arbiter; the two former drag paths
(native-only and in-app-only) are gone.

## Test plan

Unit (`./gradlew test allTests`, green):

- `WindowDragArbiterTest` — WM answers / ignores / refuses outright, release mid-wait, verdict
  surviving arbiter rebuilds, maximize mid-gesture (#76), no native support at all.
- `NativeMoveWatchTest` — settle boundaries, WM taking over, two-strikes verdict, recovery after
  the WM answers again.
- `WindowDragGestureTest` — take-over geometry against the existing dead-zone rules.

Manual, on a WM that **ignores** the request (GNOME 50 Wayland, verified):

- [x] Drag the window by empty titlebar space — it follows the pointer.
- [x] Double-click the titlebar — maximize/restore.
- [x] Click tabs and window buttons — the window does not move.
- [x] stderr carries exactly two "ignored `_NET_WM_MOVERESIZE`" lines per session, not one per drag.

Manual, on a WM that **honours** the request — not yet done:

- [ ] Drag by the titlebar — the move is compositor-driven and smooth, as before this branch.
- [ ] Drag repeatedly, ten or more times in a row — every gesture still starts a move (this is the
      path that would break if ending the gesture after the WM takes over were wrong).
- [ ] Drag, release, then immediately press the titlebar again — the second press starts a new drag.
- [ ] Double-click to maximize and restore, drag a restored window.
- [ ] Nothing is printed to stderr.

Not verified: Windows, macOS, and any non-GNOME WM.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/app/skerry/ui/desktop/NativeWindowMove.kt
+++ b/composeApp/src/desktopMain/kotlin/app/skerry/ui/desktop/NativeWindowMove.kt
@@ -17,8 +17,10 @@ import java.awt.Window
  * false and the caller keeps using the manual drag.
  */
 object NativeWindowMove {
-    // From X11/Xutil.h: the direction telling the WM to start a plain move (not a resize).
+    // From X11/Xutil.h: the direction telling the WM to start a plain move (not a resize), and the
+    // one withdrawing a request it hasn't acted on.
     private const val NET_WM_MOVERESIZE_MOVE = 8
+    private const val NET_WM_MOVERESIZE_CANCEL = 11
     // SubstructureRedirectMask (1<<20) | SubstructureNotifyMask (1<<19): the mask the WM listens on
     // for client messages posted to the root window.
     private val ROOT_EVENT_MASK = NativeLong((1L shl 20) or (1L shl 19))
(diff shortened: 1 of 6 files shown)


SeCherkasov:
Root cause found elsewhere: the failure is tied to system display scaling (anything other than 100%), not to Mutter refusing _NET_WM_MOVERESIZE per se. Closing this fallback in favour of a fix at the coordinate level.

PR #182, opened 2026-08-06
feat(ai): model catalog — refresh button, searchable picker, favorites and save hints

onepve:
The AI settings screen currently only has three fields (endpoint, API key, model) and the model name has to be typed by hand. This PR adds a model catalog flow so users can discover and pick models from the server instead:

**What's included**
- **"Refresh models" button** next to the model field — fetches the model list via `GET /models` (OpenAI-compatible `listModels()` added to `AiProvider`, with a default implementation; real providers can override). The button shows "Refreshing…" while in flight and disables itself while loading.
- **Model field becomes an editable combo** — type freely as before, or pick from the fetched catalog. The dropdown wraps the full field width (input + arrow button) so long model names aren't squeezed, and the menu caps its height with internal scrolling for catalogs with hundreds of entries.
- **Searchable picker** — the menu has a search box on top (case-insensitive substring filter) with a scrollable result list; a shared `ModelPickerMenu` component is used by both desktop and mobile.
- **Favorites (star)** — each row has a star that toggles the model as a favorite; starred models sort to the top of the list. Favorites are stored locally per server address (config-dir file on desktop, SharedPreferences on Android) and never synced.
- **Local catalog cache** — the fetched model list is cached per endpoint so the picker opens instantly without re-fetching; cached per device, not synced.
- **Save hints** — feedback next to the action buttons so the user always knows the state: editing the address/key/model shows "…changed — press Save" (amber, stays until saved); refreshing shows "Refreshing models…" then "Models refreshed" (green, auto-clears); pressing Save shows "Saved". Failures show the existing red error message.
- **Field order** — the BYOK section is now Server address → API key → Model (address first, since the other two fields talk to it).
- All new strings are added in en/zh/ru.

The picker is intentionally a simple dropdown-style menu (consistent with the rest of the settings UI) rather than a full-screen search page, to keep the settings surface compact.


onepve:
Two follow-ups on top of the model catalog:

1. **Catalog size** — the picker now shows a count line under the search field (e.g. "321 models"), so a refresh's result is visible at a glance. All three locales, positional args (%1$d) per the codebase convention.

2. **Soft-keyboard handling on mobile** — the picker opened in a Popup, which is a separate window and receives no IME insets, so the search box sat under the keyboard. The picker now expands inline in the page flow and `imePadding` sits on the scrollable column (as in the terminal view), so the keyboard lifts the page — search box and list together. Desktop has no IME and is unaffected.

SeCherkasov (commented):
## Review

Read against the code on `feat/ai-model-catalog` (base `53162cc4`, 4 commits behind `main`), not against the description. Build, tests and a device were not run.

The feature is worth having and mostly reads like the rest of the codebase. It is not mergeable yet — three things block it, and a list of per-line issues follows in the inline comments.

**1. No tests.** 13 files, +733/-32, zero test files. In this repo a feature owes a test ("RED — the failing test comes first"). Two seams are testable without any UI: `OpenAiProvider.listModels` (empty `data`, an entry without `id`, a non-JSON body, a non-2xx status) and `AiAssistantController.listModels`.

**2. The branch carries work that is not this feature.** 18 `settings_backup_*` keys x 3 locales (54 lines) belong to a data-backup feature that is not in this PR — `git grep settings_backup_ -- '*.kt'` returns nothing. Same for the `settings_security_auto_lock_desc` rewrite in all three locales, and the zh-only retranslation of `settings_ai_field_endpoint`. Please take them out of this branch.

**3. Three defects that reach beyond the feature.** The shared `DropdownMenuColumn` gets a modifier-order regression that affects every dropdown in the app; the mobile keyboard fix does not do what its comment says; and the new authenticated `GET /models` follows redirects with the `Authorization` header attached, which is a way to lose a user's API key. Details inline.

### Verified as fine, so you do not need to re-check

- All four new icon names (`star`, `star_border`, `search`, `expand_more`) exist as ligatures in `material_symbols_outlined.ttf` — checked against the font file, not assumed.
- No duplicate string keys in any of the three locales; all new AI strings are present in en/ru/zh.
- No `HttpClient` leak from the un-closed provider: `aiProviderFactory` returns `OpenAiProvider.pooled`, whose `close()` is a no-op on the shared client. The contract issue is still worth fixing — see the comment on `AiAssistantController.kt`.
- Every API the new code calls exists with the signature it is called with (`ChipButton(enabled = …)`, `AnchoredDropdown(menu = { width -> … })`, `Txt(maxLines, overflow)`, `FilePrefs`, `PrivateConfig.ensureDir`).

composeApp/src/commonMain/composeResources/values/strings_settings.xml
```diff
@@ -241,4 +253,22 @@
...
+    <string name="settings_backup_import_action">Import</string>
+    <string name="settings_backup_import_records">%1$d record(s) in this backup</string>
+    <string name="settings_backup_encrypted_note">This backup is encrypted — enter your master password to read it.</string>
+    <string name="settings_backup_merge">Merge</string>
+    <string name="settings_backup_merge_sub">Existing data stays; the backup wins where it is newer.</string>
+    <string name="settings_backup_replace">Replace</string>
+    <string name="settings_backup_replace_sub">Deletes ALL current synced data first.</string>
+    <string name="settings_backup_import_confirm">Confirm replace</string>
+    <string name="settings_backup_err_password">Wrong master password</string>
+    <string name="settings_backup_err_corrupted">This file is not a valid backup</string>
```
SeCherkasov: **Out of scope — please remove (and the ru/zh copies of the same 18 keys).**

These belong to a data-backup feature that is not part of this PR. `git grep 'settings_backup_' -- '*.kt'` finds no Kotlin usage, so they are 54 dead lines across the three locale files.

composeApp/src/commonMain/composeResources/values/strings_settings.xml
```diff
@@ -135,7 +147,7 @@
     <string name="settings_security_biometric_recheck">Check again</string>
     <string name="settings_security_biometric_weak_binding">On with a weaker key binding — biometrics still required.</string>
     <string name="settings_security_auto_lock">Auto-lock</string>
-    <string name="settings_security_auto_lock_desc">Lock after idle.</string>
+    <string name="settings_security_auto_lock_desc">Lock after idle; minimizing on desktop locks it too.</string>
```
SeCherkasov: **Out of scope — please remove (all three locales).**

This documents desktop lock-on-minimize, which has nothing to do with the model catalog. It should land with whatever change introduced that behaviour.

composeApp/src/commonMain/composeResources/values-zh/strings_settings.xml
```diff
@@ -17,14 +17,26 @@
     <string name="settings_ai_live_subtitle">AI 助手提供商与设置。</string>
     <string name="settings_ai_field_api_key">OPENAI API 密钥</string>
     <string name="settings_ai_field_model">模型</string>
-    <string name="settings_ai_field_endpoint">端点</string>
+    <string name="settings_ai_field_endpoint">服务器地址</string>
```
SeCherkasov: **Out of scope, and it makes the locales disagree.**

en is still `ENDPOINT` and ru is `АДРЕС`, while zh now reads 服务器地址 ("server address"). Either all three change or none — and either way this is not part of the model-catalog feature.

composeApp/src/commonMain/kotlin/app/skerry/ui/design/DropdownField.kt
```diff
@@ -66,21 +69,33 @@ fun SelectTrigger(value: String, onClick: () -> Unit) {
...
+fun DropdownMenuColumn(width: Dp, maxHeight: Dp = 320.dp, content: @Composable () -> Unit) {
     Column(
-        Modifier.width(width).clip(RoundedCornerShape(8.dp)).background(Skerry.colors.surface2).border(1.dp, Skerry.colors.cyan14, RoundedCornerShape(8.dp)),
+        Modifier
+            .width(width)
+            .heightIn(max = maxHeight)
+            .verticalScroll(rememberScrollState())
+            .clip(RoundedCornerShape(8.dp))
+            .background(Skerry.colors.surface2)
+            .border(1.dp, Skerry.colors.cyan14, RoundedCornerShape(8.dp)),
```
SeCherkasov: **Bug: modifier order — the surface and the border scroll away with the content.**

`verticalScroll` comes before `clip`/`background`/`border`, so everything after it is applied to the *scrolling content* (full content height, translated by the scroll offset), not to the viewport. With a menu taller than `maxHeight` the rounded border extends past the bottom of the popup and the whole frame slides as you scroll.

`ModelPickerMenu` in this same PR gets it right (`AiModelPicker.kt:69-78`): clip/background/border first, scroll on the inner column.

```kotlin
Modifier
    .width(width)
    .clip(RoundedCornerShape(8.dp))
    .background(Skerry.colors.surface2)
    .border(1.dp, Skerry.colors.cyan14, RoundedCornerShape(8.dp))
    .heightIn(max = maxHeight)
    .verticalScroll(rememberScrollState())
```

Worth flagging that this is a shared component: `DropdownField` and `TeamScreenContent.kt:243` inherit both the cap and the bug.

composeApp/src/commonMain/kotlin/app/skerry/ui/mobile/MobileAiScreen.kt
```diff
@@ -69,10 +96,15 @@ import app.skerry.ui.theme.Skerry
...
+    // imePadding on the SCROLLABLE column, not the screen root: the keyboard then both shrinks the
+    // scroll viewport and auto-scrolls the focused field (and any open inline picker) above it —
+    // mirroring the terminal view's lift. Root-level padding would only shrink the page bottom and
+    // leave a mid-page field (the model picker) covered. Popup windows receive no IME insets, which
+    // is why the picker below is inline in the page flow, not a Popup.
     Column(Modifier.fillMaxSize().background(Skerry.colors.bg)) {
         MobilePushHeader(stringResource(Res.string.more_ai_privacy), onBack = state::pop)
         if (ai == null) return@Column
-        Column(Modifier.fillMaxWidth().verticalScroll(rememberScrollState()).padding(horizontal = 18.dp)) {
+        Column(Modifier.fillMaxWidth().verticalScroll(rememberScrollState()).imePadding().padding(horizontal = 18.dp)) {
```
SeCherkasov: **The keyboard fix does not do what the comment above it claims.**

`imePadding()` placed *after* `verticalScroll()` applies to the scrolling content: it adds bottom padding inside the scroll while the container keeps its full height. The viewport is not shrunk, so a field in the middle of the page can still sit under the keyboard. The comment on lines 99-103 states the opposite ("the keyboard then both shrinks the scroll viewport").

To shrink the viewport it has to come first:

```kotlin
Column(Modifier.fillMaxWidth().imePadding().verticalScroll(rememberScrollState()).padding(horizontal = 18.dp))
```

Also worth knowing before you pick an approach: this repo already solved the same problem once. `VaultDialogs.kt:362-385` re-requests `bringIntoView` on every frame for ~450 ms and documents there why observing `WindowInsets.ime` is unreliable under `adjustResize`.

Either way this needs checking on a real device — nothing in the PR says it was.

composeApp/src/commonMain/kotlin/app/skerry/ui/settings/AiSection.kt
```diff
@@ -141,31 +169,147 @@ private fun LiveAiSection(ai: app.skerry.ui.ai.AiAssistantController) {
...
+ * editable combo: type any model id, or open the dropdown to pick one from the catalog fetched via
+ * the refresh button. Expands when the card is selected.
  */
 @Composable
 private fun DesktopByokFields(ai: app.skerry.ui.ai.AiAssistantController) {
     var key by remember(ai.settings) { mutableStateOf(ai.settings.apiKey) }
     var model by remember(ai.settings) { mutableStateOf(ai.settings.model) }
     var baseUrl by remember(ai.settings) { mutableStateOf(ai.settings.baseUrl) }
+    var models by remember(ai.settings) { mutableStateOf(AiModelCache.load(baseUrl)) }
+    var favorites by remember(ai.settings) { mutableStateOf(AiModelCache.loadFavorites(baseUrl)) }
```
SeCherkasov: **Bug: the catalog is read under one key and written under another.**

These read for `ai.settings.baseUrl` (the *saved* address, because the initialiser runs when `remember(ai.settings)` is re-keyed), while `AiModelCache.save` / `saveFavorite` on lines 263 and 300 write for the *currently typed* `baseUrl`. Change the address without pressing Save and the dropdown still lists the previous server's models, while a star you toggle is stored under the new address's key. You can then pick a model that does not exist on the endpoint you are pointing at.

Key the read to the same value you write with (`remember(baseUrl)`), or reload the cache when the field changes.

Separately: `AiModelCache.load()` runs inside composition — a file read on desktop, a `SharedPreferences` read on Android — and repeats on every `ai.settings` change. That belongs off the composition path.

composeApp/src/commonMain/kotlin/app/skerry/ui/settings/AiSection.kt
```diff
@@ -141,31 +169,147 @@ private fun LiveAiSection(ai: app.skerry.ui.ai.AiAssistantController) {
...
+                            hint = hintRefreshed; hintFlash = true
+                        },
+                        onFailure = { e ->
+                            refreshFailure = if (e is AiException) e.toFailure() else AiFailure.UNKNOWN
+                            hint = null
+                        },
+                    )
+                }
+            },
+            enabled = key.isNotBlank() && baseUrl.isNotBlank() && !refreshing,
```
SeCherkasov: **The refresh button is disabled for exactly the setups that need it most.**

`key.isNotBlank()` rules out Ollama, LM Studio and any local proxy, none of which take an API key — and the app deliberately supports those (`isInsecureAiEndpoint` exempts localhost from the cleartext warning for that reason). `baseUrl.isNotBlank() && !refreshing` is enough. Same line in `MobileAiScreen.kt:309`.

composeApp/src/commonMain/kotlin/app/skerry/ui/settings/AiSection.kt
```diff
@@ -141,31 +169,147 @@ private fun LiveAiSection(ai: app.skerry.ui.ai.AiAssistantController) {
...
+            Row(horizontalArrangement = Arrangement.spacedBy(6.dp), verticalAlignment = Alignment.CenterVertically) {
+                Box(Modifier.weight(1f)) {
+                    SyncField(placeholder = stringResource(Res.string.settings_ai_placeholder_model), value = model, icon = "auto_awesome", keyboardType = KeyboardType.Text, imeAction = ImeAction.Done, onChange = { model = it })
+                }
+                Box(
+                    Modifier
+                        .clip(RoundedCornerShape(7.dp))
+                        .background(Skerry.colors.bg)
+                        .border(1.dp, Skerry.colors.cyan14, RoundedCornerShape(7.dp))
+                        .clickable(enabled = models.isNotEmpty()) { modelMenuOpen = !modelMenuOpen }
```
SeCherkasov: **Silent no-op before the first refresh.**

`clickable(enabled = models.isNotEmpty())` with no disabled styling: the arrow looks live and does nothing until a refresh has populated the cache, with no feedback at all. Either grey it out, or open the menu regardless and let it show the "No models found" text it already handles. Same at `MobileAiScreen.kt:244`.

composeApp/src/commonMain/kotlin/app/skerry/ui/settings/AiSection.kt
```diff
@@ -141,31 +169,147 @@ private fun LiveAiSection(ai: app.skerry.ui.ai.AiAssistantController) {
...
 /**
- * BYOK fields inside the "My API key" card: key/model/endpoint + Save (encryption happens in
- * [app.skerry.ui.ai.AiAssistantController.save]). Expands when the card is selected.
+ * BYOK fields inside the "My API key" card: endpoint / API key / model + Refresh models + Save
+ * (encryption happens in [app.skerry.ui.ai.AiAssistantController.save]). The model field is an
+ * editable combo: type any model id, or open the dropdown to pick one from the catalog fetched via
+ * the refresh button. Expands when the card is selected.
  */
 @Composable
 private fun DesktopByokFields(ai: app.skerry.ui.ai.AiAssistantController) {
```
SeCherkasov, line 172: **~55 lines duplicated verbatim between desktop and mobile.**

`DesktopByokFields` and `MobileByokFields` (`MobileAiScreen.kt:173`) now share: the seven state holders, the hint `LaunchedEffect`, the six hoisted `stringResource` hints, the favourite toggle, and the whole refresh handler including the `Result.fold`. Only the visual part was extracted (`ModelPickerMenu`).

Pull the state out into commonMain — e.g. `rememberByokModelState(ai)` exposing the states plus `refresh()` and `toggleFavorite(id)` — so each screen keeps only its layout. `docs/coding-guidelines.md` §1: a second copy is the point at which to extract.

composeApp/src/commonMain/kotlin/app/skerry/ui/mobile/MobileAiScreen.kt
```diff
@@ -133,35 +165,151 @@ fun MobileAiScreen(state: MobileDesignState) {
...
-        MobileFormField(stringResource(Res.string.settings_ai_field_endpoint)) {
-            MobileFormInput(baseUrl, { baseUrl = it }, stringResource(Res.string.settings_ai_placeholder_endpoint), keyboardType = KeyboardType.Uri, imeAction = ImeAction.Done)
+        // The picker expands IN the page flow instead of a Popup: Popup windows get no IME insets and
+        // stay put under the soft keyboard, but the screen root's imePadding lifts this whole page —
+        // search box and list together — above the keyboard when the search field gains focus. Desktop
+        // has no IME: the inline menu simply pushes the Save row down (scrollable, fine).
+        if (modelMenuOpen) {
+            Spacer(Modifier.height(8.dp))
+            ModelPickerMenu(
+                width = Dp.Infinity,
```
SeCherkasov: **`Dp.Infinity` in `Modifier.width()` does not mean "full width".**

`Dp.Infinity` is intended for `*In(max = …)` bounds. Passed to `Modifier.width()` it collapses to min 0 / max ∞ and the menu ends up sized by its content, bounded by the parent — the opposite of the "wraps the full field width" claim in the PR description.

Give `ModelPickerMenu` a `Modifier` parameter (or a `fillMaxWidth` flag) instead of a `Dp`, and pass `Modifier.fillMaxWidth()` here while desktop passes `Modifier.width(anchorWidth)`.

composeApp/src/commonMain/kotlin/app/skerry/ui/ai/AiModelPicker.kt
```diff
@@ -0,0 +1,160 @@
...
+        ) {
+            if (filtered.isEmpty()) {
+                Txt(
+                    if (models.isEmpty()) emptyText else stringResource(Res.string.settings_ai_no_matches),
+                    color = Skerry.colors.faint,
+                    size = 12.5.sp,
+                    modifier = Modifier.padding(horizontal = 12.dp, vertical = 10.dp),
+                )
+            } else {
+                filtered.forEach { m ->
```
SeCherkasov: **Non-lazy list for a catalog the PR itself describes as hundreds of entries.**

`Column(verticalScroll) { filtered.forEach { … } }` composes every row up front; the menu even prints "321 models" from `settings_ai_models_count`. This should be a `LazyColumn`.

composeApp/src/commonMain/kotlin/app/skerry/ui/ai/AiModelPicker.kt
```diff
@@ -0,0 +1,160 @@
...
+            stringResource(Res.string.settings_ai_models_count, models.size),
+            color = Skerry.colors.faint,
+            size = 11.sp,
+            modifier = Modifier.padding(horizontal = 12.dp, vertical = 5.dp),
+        )
+        HLine()
+        Column(
+            Modifier
+                .fillMaxWidth()
+                .heightIn(max = (maxHeight - 40.dp).coerceAtLeast(120.dp))
```
SeCherkasov: **The 40.dp constant does not match the header, so the bottom of the list gets clipped.**

Above this column sit the search row (~30dp with its 8dp vertical padding), the count line (~21dp) and an `HLine` — more than 40dp. The inner cap plus the header therefore exceeds `maxHeight`, and the outer `heightIn(max = maxHeight)` on line 73 cuts off the end of the list.

Drop the arithmetic and let layout do it: give this column `Modifier.weight(1f)` inside the capped outer `Column`.

composeApp/src/commonMain/kotlin/app/skerry/ui/ai/AiAssistantController.kt
```diff
@@ -154,6 +155,20 @@ class AiAssistantController(
...
 
+    /**
+     * Fetches the model catalog for the BYOK fields (refresh button in AI settings), using the
+     * typed key/baseUrl. The catalog is not persisted — the UI keeps it per screen. Failures are
+     * surfaced as [Result.failure] (an [AiException] when the endpoint answered, anything else
+     * otherwise); the UI maps them to a localized message.
+     */
+    suspend fun listModels(apiKey: String, baseUrl: String): Result<List<String>> = runCatching {
+        providerFactory(
+            AiEndpoint.Cloud(
```
SeCherkasov: **`runCatching` swallows `CancellationException` — the one thing this codebase says never to swallow.**

`OpenAiProvider.chat` states the rule explicitly: *"coroutine cancellation is not a provider failure — never swallow or rewrap it"*. Here a cancelled refresh (the screen leaves composition mid-flight) comes back as `Result.failure`, and the caller then goes on to write `refreshing = false` and `refreshFailure = AiFailure.UNKNOWN` from a cancelled coroutine. Catch `CancellationException` and rethrow before wrapping anything else.

Second point on the same block: the provider created here is never closed, unlike `AiStreamRunner`, which always does `provider.close()` in a `finally`. It is harmless today because `aiProviderFactory` hands back `OpenAiProvider.pooled` (whose `close()` is a no-op on the shared client), but the interface's contract says otherwise and the day the factory returns an owning provider this leaks a CIO client per button press. Wrap it in `try { … } finally { provider.close() }`.

shared/src/jvmSharedMain/kotlin/app/skerry/shared/ai/OpenAiProvider.kt
```diff
@@ -98,6 +100,23 @@ class OpenAiProvider private constructor(
         return chunk.choices.firstOrNull()?.delta?.content
     }
 
+    /**
+     * Fetches the model catalog (`GET {baseUrl}/models`) for the BYOK settings refresh button. The
+     * key is only used in this request's `Authorization` header, never logged.
+     */
+    override suspend fun listModels(): List<String> {
+        val response = http.get("${config.baseUrl}/models") {
```
SeCherkasov: **The user's API key can be redirected to an arbitrary host.**

This is the first authenticated `GET` in the client, and that matters: Ktor installs `HttpRedirect` by default (`followRedirects = true`) and follows 3xx for GET/HEAD only — which is why `chat`'s POST was never redirected. On a redirect Ktor replays the original headers, `Authorization` included, against whatever host `Location` names. A compromised endpoint, or a MITM on an `http://` endpoint (which the UI permits with a warning), harvests the key with a single 302.

Fix by not following redirects on this call — either a client configured with `followRedirects = false` for the catalog request, or treat a 3xx status as a failure explicitly.

shared/src/jvmSharedMain/kotlin/app/skerry/shared/ai/OpenAiProvider.kt
```diff
@@ -98,6 +100,23 @@ class OpenAiProvider private constructor(
...
+            header(HttpHeaders.Authorization, "Bearer ${config.apiKey}")
+            header(HttpHeaders.Accept, ContentType.Application.Json.toString())
+        }
+        if (!response.status.isSuccess()) throw errorFor(response.status)
+        return try {
+            json.decodeFromString(ModelsWire.serializer(), response.bodyAsText()).data.map { it.id }
```
SeCherkasov: **Untrusted response read whole, with no cap on either the body or the catalog.**

`bodyAsText()` pulls the entire response into memory, and the resulting list is unbounded. A broken or hostile endpoint returning a huge `/models` document lands it in memory, then in the desktop config file / Android `SharedPreferences`, then in a non-lazy `Column` (`AiModelPicker.kt:130`). `chat` never had this exposure because it streams line by line.

Cap the number of ids (a couple of thousand is far past any real catalog) and the length of a single id, and drop the rest.

composeApp/src/androidMain/kotlin/app/skerry/ui/AiModelCache.android.kt
```diff
@@ -0,0 +1,34 @@
...
+actual object AiModelCache {
+
+    private val prefs: android.content.SharedPreferences?
+        get() = SafBridge.context()?.getSharedPreferences("ai_model_cache", Context.MODE_PRIVATE)
+
+    actual fun load(baseUrl: String): List<String> =
+        prefs?.getString(cacheKey(baseUrl), null)?.lines()?.filter { it.isNotBlank() } ?: emptyList()
+
+    actual fun save(baseUrl: String, models: List<String>) {
+        prefs?.edit()?.putString(cacheKey(baseUrl), models.joinToString("\n"))?.apply()
```
SeCherkasov: **A model id containing a newline corrupts the cache — and the two platforms disagree about it.**

`joinToString("\n")` is not filtered, so a server-supplied id with `\n` in it splits into two entries on the next `load()`. The desktop actual drops such values (`FilePrefs.setLines` filters `\n` and `\r`, for exactly this reason). Filter here too, or better, share one sanitising step in `commonMain`.

composeApp/src/androidMain/kotlin/app/skerry/ui/AiModelCache.android.kt
```diff
@@ -0,0 +1,34 @@
...
+        if (favorite) fav.add(id) else fav.remove(id)
+        prefs?.edit()?.putString(favKey(baseUrl), fav.joinToString("\n"))?.apply()
+    }
+
+    /** Stable per-address key (16-hex of the trimmed address hash; no URL chars in prefs keys). */
+    private fun cacheKey(baseUrl: String): String =
+        "mc_" + baseUrl.trim().hashCode().toUInt().toString(16)
+
+    private fun favKey(baseUrl: String): String =
+        "mf_" + baseUrl.trim().hashCode().toUInt().toString(16)
```
SeCherkasov: **32-bit `hashCode()` as a per-address cache key; the desktop actual of the same class uses SHA-1.**

Two different endpoints can collide and swap catalogs and favourites. Use one scheme on both platforms — SHA-1 truncated the way `AiModelCache.desktop.kt` already does it.

composeApp/src/desktopMain/kotlin/app/skerry/ui/AiModelCache.desktop.kt
```diff
@@ -0,0 +1,46 @@
...
+            .joinToString("") { it.toInt().and(0xff).toString(16).padStart(2, '0') }
+        return "ai_model_cache_" + digest.take(12)
+    }
+
+    private fun favKey(baseUrl: String): String {
+        val digest = MessageDigest.getInstance("SHA-1")
+            .digest(baseUrl.trim().toByteArray(Charsets.UTF_8))
+            .joinToString("") { it.toInt().and(0xff).toString(16).padStart(2, '0') }
+        return "ai_model_fav_" + digest.take(12)
+    }
```
SeCherkasov: **The same SHA-1 block written twice, differing only in the prefix.**

Extract it:

```kotlin
private fun digest(baseUrl: String): String =
    MessageDigest.getInstance("SHA-1")
        .digest(baseUrl.trim().toByteArray(Charsets.UTF_8))
        .joinToString("") { it.toInt().and(0xff).toString(16).padStart(2, '0') }
        .take(12)

private fun cacheKey(baseUrl: String) = "ai_model_cache_" + digest(baseUrl)
private fun favKey(baseUrl: String) = "ai_model_fav_" + digest(baseUrl)
```

composeApp/src/desktopMain/kotlin/app/skerry/ui/AiModelCache.desktop.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    private val dir: Path = configDir()
+    private val prefs = FilePrefs(dir)
+
+    /** `~/.config/skerry` (or $XDG_CONFIG_HOME), hardened — mirrors the app's main config dir. */
+    private fun configDir(): Path {
+        val xdg = System.getenv("XDG_CONFIG_HOME")?.takeIf { it.isNotBlank() }
+        val base = xdg?.let { Path.of(it) } ?: Path.of(System.getProperty("user.home"), ".config")
+        return base.resolve("skerry").also { PrivateConfig.ensureDir(it) }
+    }
```
SeCherkasov: **Verbatim copy of an existing function.**

This is `configDir()` from `composeApp/src/desktopMain/kotlin/app/skerry/ui/main.kt:106`, character for character, `PrivateConfig.ensureDir` included. Make that one `internal` and call it instead — `docs/coding-guidelines.md` §1 asks for a search for an existing abstraction before a new one, and this is a literal duplicate.

onepve:
All review points addressed in `72f9e053` (push just now, CI will re-run on the PR).

One deviation worth flagging on the `settings_ai_field_endpoint` note: instead of reverting that string, we aligned all three locales on the common user-facing term for this field — zh 接口地址 (the standard label for the OpenAI-compatible base URL in Chinese clients such as NextChat / DeepSeek's docs / LobeChat), en "API URL", ru "API URL". The locale mismatch you pointed out is gone; the label is just less jargon-y than "ENDPOINT". It rides along with this PR as a wording fix — if you'd rather keep the original technical label or land it separately, we'll take it out.

SeCherkasov:
Thanks for the feature, and for working through the whole review! I pushed a few small changes of my own on top to close the remaining tails.

PR #216, opened 2026-08-07
Add Maple Mono font (CJK-friendly monospace, OFL-1.1)

onepve:
This adds Maple Mono (regular + bold) as bundled font resources for CJK-friendly terminal rendering, plus its OFL-1.1 license text.

Maple Mono is a round-cornered monospace with strict 2:1 width alignment between CJK glyphs and ASCII — handy when Chinese/Japanese text appears in terminal output.

Original font (OFL-1.1): https://github.com/subframe7536/maple-font

The bundled weights are subset to GB2312 + ASCII + CJK punctuation (~7605 glyphs, ~5 MB each) to keep the app lean; the full font covers all of CJK (~18 MB per weight) — grab it from the upstream repo if you want the complete set.

Measured on our Android build: the two subset weights add **~9.7 MB** to the final APK (82.5 MB → 92.2 MB).

No code changes — just the font resources, for reference.


SeCherkasov:
Thanks — the font itself checks out, so I pushed a follow-up commit to this branch instead of asking you to redo it. Your description stays as it is; here is what the second commit changes.

**Wired the font up.** `TerminalFont` only offered Hack and JetBrains Mono, so the two 5 MB faces shipped in every artifact with no way to select them. Maple Mono CN is now the third entry in Appearance → Font (`id = "maple-mono"`), and both pickers — desktop `TerminalSection.kt` and `MobileAppearanceScreen.kt` — iterate `entries`, so they pick it up on their own. Tests: a new `a_cjk_capable_font_is_offered` plus an id-uniqueness check in `TerminalAppearanceTest`.

**Moved the license out of `composeResources/font/`.** Anything in that directory becomes a `FontResource`, and the OFL text was no exception:

```kotlin
internal val Res.font.MAPLEMONO_LICENSE: FontResource by lazy {
  FontResource("font:MAPLEMONO_LICENSE", ... "${MD}font/MAPLEMONO-LICENSE.txt" ...)
```

It now lives in `licenses/MapleMono-OFL-1.1.txt`. While there, I added the texts for the four fonts that were already bundled with no notice at all — Hack, JetBrains Mono, IBM Plex, Material Symbols — with `licenses/README.md` listing versions and upstreams, linked from both READMEs. That gap was ours, not yours; your PR just made it visible.

**Documented the subset.** Nothing in the repo proved what the binaries contain, so `tools/fonts/verify-maple-mono.py` checks them against upstream `v7.9`: every bundled glyph must exist there with the same outline and advance, and CJK must stay exactly 2:1 against ASCII. Both weights pass — 7426 codepoints, all verified. The repertoire is `(GB2312 ∪ ASCII ∪ U+3000–303F) ∩ upstream`, and since the Maple Mono copyright carries no Reserved Font Name, keeping the family name on a subset is fine under OFL-1.1.

PR #254, opened 2026-08-15
feat(android): keep SSH sessions alive in the background with per-session notifications

onepve:
## Problem

Found in normal daily use: when running long commands on multiple terminals and switching to another app, Android freezes the app's network state, the SSH connection drops, and the work is lost. With keep-alive, backgrounding the app keeps the connections alive, so several terminals can keep running long tasks (builds, deploys, transfers) in parallel while you do something else — and you come back to the same live terminal.

## What this does

- A **foreground service** (`connectedDevice` type — no 6-hour timeout, and its semantics fit: keeping a network connection to an external device alive) runs while at least one session is open, and stops itself when the last session closes. The SSH keep-alive pings keep flowing in the background, so NAT/firewall idle timers don't fire and the session doesn't drop.
- **One notification per live session** (Termius-style): a summary notification shows the active count; each session gets its own notification titled with the host. Tapping one routes back to that exact terminal tab (the session id travels in the intent).
- **Notifications re-show if swiped away** (some OEMs allow swiping ongoing FGS notifications); the keep-alive state stays visible as long as sessions are alive.
- `KeepAliveNotificationRegistry` is pure Kotlin with **7 unit tests** covering id allocation and the idempotent lifecycle.
- `MobileDesignApp` gains an optional `processScope` (null → composition scope); **desktop behavior is unchanged** — the shared bridge hook is a no-op when not set.

## Security

Tapping a notification only opens the app and activates the corresponding terminal — it does **not** bypass the master-password/biometric lock; the vault still requires unlocking as usual.

This software is really great to use, I hope it keeps getting better. Thanks for your work!

### DIFF
--- a/androidApp/src/main/AndroidManifest.xml
+++ b/androidApp/src/main/AndroidManifest.xml
@@ -9,6 +9,18 @@
     <!-- USB-OTG serial (CDC/FTDI/CP210x/CH34x adapters). Not mandatory: we do not cut off devices
          without USB Host — serial will simply show no ports. -->
     <uses-feature android:name="android.hardware.usb.host" android:required="false" />
+    <!-- Foreground service while an SSH session is open (keep-alive in the background). connectedDevice
+         type: no timeout, and keeping a network connection to an external device alive is its purpose. -->
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_CONNECTED_DEVICE" />
+    <!-- Required by targetSDK 36 to START a connectedDevice FGS: at least one of the anyOf set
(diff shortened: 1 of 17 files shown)


SeCherkasov:
Thanks — the feature direction is right and the service/notification design survived review. It did need substantial fixes, which I've pushed to your branch (your commit is preserved): the drop path used to stop the service before auto-reconnect, and a background FGS restart is forbidden on 12+, so the exact scenario this feature targets killed its own reconnect; the bridge and the session graph are now process-scoped correctly (the per-Activity rebuild lost track of sessions and pinned the first Activity's vault — a silent data-loss hazard); tap routing now works for split panes and is nonce-guarded against foreign intents; host names are redacted on the lock screen and sanitised; plus lint fixes and 13 behavioural tests on the bridge contract. See a3c35453 for the full story. Not yet verified on a device — that's next.

PR #339, opened 2026-08-26
Treat a remote listing as the server's answer, not a fact (#309, #306)

### LINKED ISSUE #306: SFTP transfer: the tree walk is unbounded, and the StackOverflowError it produces escapes the transfer's error handling
## What happens

`DownloadWalk.walk` / `UploadWalk.walk` in `composeApp/src/commonMain/kotlin/app/skerry/ui/files/TransferPlan.kt` recurse once per directory level with no depth cap, no visited set and no cap on the plan size:

```kotlin
FileItemType.Directory -> {
    plan.dirs += localPath
    sftp.list(remotePath).forEach { child ->
        walk(child.name, childPath(remotePath, child.name), child.type.toItemType(), child.size, localPath)
    }
}
```

If the remote tree loops — a bind mount that contains itself, a FUSE mount, or a hostile SFTP server that simply lists a directory as its own child — the walk never terminates. It grows `plan.dirs`/`plan.files` until the heap gives out, or blows the JVM stack first, whichever comes first.

Symlinks are not the vector: `FileItemType.Symlink` is skipped. Directory-typed loops are.

## Why it is worse than a hang

`buildDownloadPlan` is called from `runDownload` inside `TransferCoordinator.launchExclusive`, whose handler is:

```kotlin
} catch (e: CancellationException) {
    transfers.end(TransferStatus.Failed(FileTransferFailure.Transfer)); throw e
} catch (e: Exception) {
    val failure = (e as? FileBrowserException)?.failure?.toTransferFailure() ?: FileTransferFailure.Transfer
    transfers.end(TransferStatus.Failed(failure))
} finally { onFinally(); busy = false }
```

`StackOverflowError` is an `Error`, not an `Exception`, so it matches neither clause. `transfers.end` is never called and the queue keeps an entry that stays Active forever; the `finally` clears `busy`, so the UI accepts a new transfer while the old one is still shown as running. `OutOfMemoryError` behaves the same way.

The whole plan is also built before the first byte moves, so even a legitimately huge tree gives an Active entry with no progress for as long as the walk takes.

## Expected

- Bound the walk: a maximum depth, a cap on the number of entries in a plan, or both — and fail with a typed `FileBrowserFailure` when exceeded, the way `isUnsafeListingName` already fails on a bad name.
- Widen the `launchExclusive` handler so a `Throwable` that is not `CancellationException` still ends the queue entry as Failed instead of leaving it Active.

## Repro

A remote directory that lists itself as a child is enough; a fake `SftpClient` in `composeApp/src/commonTest/.../TransferPlanTest.kt` reproduces it without a server.

### LINKED ISSUE #309: SFTP panel: a directory listing with two identical names crashes composition (duplicate LazyColumn key)
## What happens

`composeApp/src/commonMain/kotlin/app/skerry/ui/sftp/SftpListing.kt:102` keys the file list by the path the server reported:

```kotlin
items(entries, key = { it.path }) { entry ->
```

Nothing on the way there guarantees the paths are distinct. `SftpFileBrowser.list` is a straight map:

```kotlin
override suspend fun list(path: String): List<FileItem> =
    guard { sftp.list(path).map { it.toFileItem() } }
```

and `SshjSftpClient.list` is `sftp.ls(path).map { it.toEntry() }` — sshj hands back whatever the server sent, minus `.` and `..`. Two entries with the same name in one directory produce two `FileItem`s with the same `path`.

## Why it crashes

`LazyColumn` wraps every item in `SaveableStateProvider(key)`, and `SaveableStateHolder` (compose-runtime-saveable 1.9.4, `SaveableStateHolder.kt:88`) refuses a duplicate:

```kotlin
require(key !in registries) { "Key $key was used multiple times " }
```

Duplicated names land as adjacent rows, so both are composed in the same pass — the `IllegalArgumentException` is thrown during composition and takes the window (desktop) or the activity (Android) with it. Opening a directory is enough; there is no way back into the panel without a restart.

An SFTP server the client does not trust is inside this project's threat model, and a listing with a repeated filename is one packet away. A merged/overlay filesystem or a buggy server reaches the same state without malice.

## Elsewhere in the codebase this is already handled

`OpenAiProvider.listModels` de-duplicates the catalog before it reaches `AiModelPicker`'s `items(filtered, key = { it })`, with the reason spelled out in a comment (`… or the desktop cache … would return duplicates`), and `commandSuggestions` de-duplicates through a `LinkedHashMap`. The SFTP listing is the one untrusted list that does not.

## Expected

De-duplicate in `SftpFileBrowser.list` (keep the first of each path, the way the model catalog keeps `.distinct()`), or key the rows by index. First is better: a duplicate name is a broken listing, and showing it twice is wrong regardless of the crash.

## Test

A fake `SftpClient` returning two entries with the same name, listed through `SftpFileBrowser`, plus a Compose test that renders the panel over that listing.

SeCherkasov:
Two issues about the same thing: a directory listing is the server's answer, not a fact, and the file
panel treated it as one. Two commits, one per issue — **merge without squashing**.

## `SftpFileBrowser.list` de-duplicates a listing (#309)

Nothing in SFTP promises the entries in one listing are distinct. An overlay/merged filesystem
repeats a name by construction; a buggy or hostile server does it by choice. Both halves of an entry
have to be unique, for two different reasons:

- the panel keys its rows by the reported path, and compose-runtime-saveable refuses a duplicate key
  with an `IllegalArgumentException` thrown *during composition* — it takes the window down on
  desktop and the activity on Android, and re-opening the directory hits it again;
- a row draws the name and never the path, while every operation on a row resolves that name against
  the directory it was listed in, so two rows under one name are two rows the user cannot tell apart
  that act on one file — deleting the second deletes what the first one showed.

The first entry of each path and of each name wins: which the server meant is unknowable, and
keeping the later one would make the listing depend on packet order. `FileBrowser.list` states
distinctness as part of the contract. The download walk plans from the SFTP client rather than from
that browser, so it de-duplicates its own child listing by name.

## The tree walks are bounded, and a dead operation closes its row (#306)

`DownloadWalk`/`UploadWalk` recursed once per directory level with no cap on depth and none on the
size of the plan. A tree that loops — a bind mount that contains itself, a FUSE mount, or a server
that lists a directory as its own child — never terminates: the walk blows the stack before a byte
has moved. Symlinks were never the vector; directory-typed loops are.

- Both walks run under a `TreeWalkLimit` (64 levels, 100 000 entries) and refuse what exceeds it
  with a typed `FileBrowserFailure.TreeTooLarge`, so the queue row says why.
- The recursive SFTP delete walks the same untrusted listing on the more destructive verb and takes
  the **depth** half of that bound only. It deletes as it goes rather than deciding first, so an
  entry cap would stop it with most of the tree already gone — on a verb where "refused" and "half
  done" are not the same answer.
- `StackOverflowError` is an `Error`, so `TransferRunner`'s `catch (e: Exception)` never matched it
  and the queue entry stayed Active for good: a progress bar that never moves again, and an idle
  auto-lock that keeps deferring on it. It catches `Throwable` now and closes the entry as Failed.
  `FilePaneController` and the viewer/editor do the same on their own paths — both run on the scope
  the whole composition shares, where an uncaught throw cancels every other session's work and, on
  Android, kills the process.
- A pane operation that ends in a failure is now announced. The listing and the failure notice are
  two branches of one `when`, so the switch between them is an insertion rather than a change a
  screen reader is told about; a user who confirmed a delete and lost it heard nothing.

## Behaviour change worth naming

A transfer of a tree with more than 100 000 entries is now **refused** rather than attempted. The
plan is held whole in memory before the first byte moves, so the cap is what the client can hold —
not an opinion about how much a person may transfer. The recursive delete is not capped this way and
still deletes trees of any width.

## Verification

`checks · tests · build (lint) · detekt · :androidApp:compileDebugKotlin` all green, and the first
commit compiles and passes its own tests in isolation, so bisecting through the pair works.

By hand, on Linux desktop: open a directory whose listing repeats an entry — before, the window
went; now one row is drawn. Start a download of a directory that lists itself — before, the app
died on a stack overflow; now the queue row reads "Folder tree is too deep or too large to
transfer" and the next transfer gets its turn.

**Not verified:** Android, a live server, TalkBack/Orca/NVDA, Windows and macOS.

## Known, deliberately out of scope

A single directory listing is still unbounded at `SshjSftpClient.list` (`sftp.ls`), so the entry cap
and the de-duplication both sit above the allocation they bound. A hostile server can still drive
the heap during an ordinary directory open — now recovered as a failed pane rather than a dead
process. The fix is a sshj `RemoteResourceSelector` returning `BREAK` past a limit; it belongs to
its own issue, not to either of these.

Closes #309
Closes #306

### DIFF
--- a/composeApp/src/commonMain/composeResources/values-ru/strings_ftail.xml
+++ b/composeApp/src/commonMain/composeResources/values-ru/strings_ftail.xml
@@ -28,6 +28,9 @@
     <string name="ftail_err_open_target">Не удалось открыть файл для записи</string>
     <string name="ftail_err_session_closed">Сессия закрыта до начала передачи</string>
     <string name="ftail_err_too_large">Файл слишком большой, чтобы открыть его</string>
+    <string name="ftail_err_tree_too_large">Дерево каталогов слишком глубокое или слишком большое для передачи</string>
+    <string name="ftail_err_tree_too_deep">Дерево каталогов слишком глубокое для обхода</string>
+    <string name="ftail_err_unexpected">Непредвиденная ошибка</string>
 
     <!-- Ошибки просмотрщика/редактора (F3/F4) -->
     <string name="ftail_edit_err_read">Не удалось прочитать файл</string>
--- a/composeApp/src/commonMain/composeResources/values-zh/strings_ftail.xml
+++ b/composeApp/src/commonMain/composeResources/values-zh/strings_ftail.xml
@@ -28,6 +28,9 @@
     <string name="ftail_err_open_target">无法打开保存目标</string>
     <string name="ftail_err_session_closed">会话在传输开始前已关闭</string>
     <string name="ftail_err_too_large">文件太大，无法打开</string>
+    <string name="ftail_err_tree_too_large">目录树太深或太大，无法传输</string>
+    <string name="ftail_err_tree_too_deep">目录树太深，无法遍历</string>
+    <string name="ftail_err_unexpected">意外错误</string>
 
     <!-- 查看器/编辑器（F3/F4）失败原因 -->
     <string name="ftail_edit_err_read">读取文件失败</string>
--- a/composeApp/src/commonMain/composeResources/values/strings_ftail.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings_ftail.xml
@@ -28,6 +28,9 @@
     <string name="ftail_err_open_target">Failed to open the save target</string>
     <string name="ftail_err_session_closed">Session closed before it started</string>
     <string name="ftail_err_too_large">File is too large to open</string>
+    <string name="ftail_err_tree_too_large">Folder tree is too deep or too large to transfer</string>
+    <string name="ftail_err_tree_too_deep">Folder tree is too deep to walk</string>
+    <string name="ftail_err_unexpected">Unexpected error</string>
 
     <!-- Viewer/editor (F3/F4) failures -->
     <string name="ftail_edit_err_read">Failed to read the file</string>
--- a/composeApp/src/commonMain/kotlin/app/skerry/ui/files/FileEditController.kt
+++ b/composeApp/src/commonMain/kotlin/app/skerry/ui/files/FileEditController.kt
@@ -156,9 +156,10 @@ class FileEditController(
                 state = FileEditState.Failed(
                     if (e.failure == FileBrowserFailure.TooLarge) FileEditFailure.TooLarge else FileEditFailure.Read,
                 )
-            } catch (_: Exception) {
+            } catch (@Suppress("TooGenericExceptionCaught") _: Throwable) {
                 // Anything the source didn't wrap still has to land somewhere: an unhandled type
-                // would otherwise leave the editor spinning on Loading with no explanation.
+                // would otherwise leave the editor spinning on Loading with no explanation, and an
+                // Error would take the session's shared scope down on its way out.
                 state = FileEditState.Failed(FileEditFailure.Read)
             } finally {
                 busy = false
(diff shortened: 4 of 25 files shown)


SeCherkasov:
The unbounded single listing named under "Known, deliberately out of scope" is now #340.

PR #349, opened 2026-08-30
feat(ui): add manual folder drag-and-drop reordering for snippets and runbooks, and enhance folder drag UX

onepve:
### Summary
This PR brings manual drag-and-drop folder reordering to Snippets and Runbooks (matching the existing Host sidebar capability) and improves the overall folder dragging UX across desktop and mobile platforms:

1. **First-appearance folder order**: Replaced the rigid alphabetical sorting in `foldersOf()` with natural first-appearance order, allowing user-defined drag-and-drop reordering across all foldered sections.
2. **Folder reordering for Snippets & Runbooks**: Added `moveSnippetGroup` and `moveRunbookGroup` with pure reordering algorithms and full unit test coverage.
3. **Compact drag UX**: When dragging any folder, all folder sections temporarily collapse into compact headers during the gesture, eliminating large vertical jumps and making reordering fast and predictable. Folders instantly restore their previous expansion states upon release.
4. **Touch & Click gesture conflict resolution**: Isolated fold click targets to the chevron/text elements, preventing row-level click handlers from intercepting mouse drags or touch long-presses.
5. **Drop indicator enhancement**: Improved `DropLine` visibility with a 3px glowing cyan pill and anchor dots for clear visual feedback on both high-DPI desktop screens and mobile displays.

Discovered and refined during normal daily use across desktop and Android devices.

### DIFF
--- a/composeApp/src/commonMain/kotlin/app/skerry/ui/design/FolderSectionHeader.kt
+++ b/composeApp/src/commonMain/kotlin/app/skerry/ui/design/FolderSectionHeader.kt
@@ -1,20 +1,25 @@
 package app.skerry.ui.design
 
 import androidx.compose.foundation.background
+import androidx.compose.foundation.border
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.key
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.semantics.stateDescription
@@ -26,22 +31,15 @@ import androidx.compose.ui.unit.sp
 import app.skerry.ui.generated.resources.Res
 import app.skerry.ui.generated.resources.shtail_group_collapse
 import app.skerry.ui.generated.resources.shtail_group_expand
+import app.skerry.ui.generated.resources.shtail_group_rename
 import app.skerry.ui.generated.resources.shtail_group_state_collapsed
 import app.skerry.ui.generated.resources.shtail_group_state_expanded
 import app.skerry.ui.theme.Skerry
 import org.jetbrains.compose.resources.stringResource
 
 /**
- * Folder header of a list section: chevron + folder icon + name + count, the whole row folding the
- * section on a click.
- *
- * The host sidebar's own header ([app.skerry.ui.terminal] `FolderHeader`) stays separate because
- * there the row is a drag handle and a drop target, so only its chevron may take the click. Here
- * nothing is dragged, and the full-width row is the target — which is also what makes it usable on
- * a phone, where a 22dp chevron is not.
- *
- * The row names itself after what a click does and to which folder: a screen of headers all saying
- * "Collapse" says nothing about which one is which.
+ * Folder header of a list section: chevron + folder icon + name + count.
+ * The chevron toggles collapsed state; the header body acts as a drag surface.
  */
 @Composable
 fun FolderSectionHeader(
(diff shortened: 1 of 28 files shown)


SeCherkasov:
Thanks — this is a big one, so I read the whole diff against our tree rather than the summary. Short version: **the headline feature is not reachable from the app, and several things that do ship are regressions.** Details below, with what I'd like at the end.

## 1. The drag-and-drop reordering is not wired to any UI

`FolderSections` gained `onMoveItem`, `onMoveItems`, `onMoveGroup`, `onEditGroup`, `selectedIds`, `longPress`. Not one call site passes any of them:

| Call site | new callbacks passed |
|---|---|
| `SnippetsView.kt:145` | none |
| `MobileSnippetLibrary.kt:87` | none |
| `RunbooksView.kt:152` | none |
| `MobileRunbooksView.kt:144` | none |
| `VaultView.kt:193`, `MobileVaultView.kt:250` | none |

So `canMove` and `canMoveGroup` are always false, no drag modifier is ever attached, `ListDragState` never starts, and `isAnyFolderDragging` is always false. The whole of `FolderSectionsDnd.kt` (346 new lines) is dead at runtime.

Same for the manager API: `SnippetManager.moveSnippet/moveSnippets/moveGroup/renameGroup/deleteGroup` and the `RunbookManager` equivalents have callers **only in tests**. `moveSnippetsToGroup`, `moveSnippetToGroup`, `renameSnippetGroup` and the runbook twins have no callers at all. The runbook grouping module adds `RunbookCategory`, `groupRunbooksByCategory`, `hasRunbookCategories`, `runbookTags`, `runbookCategoryChips`, `runbookGroupChips`, `filterRunbooks`, `runbookChipLabel`, `runbookGroupChipLabel` — zero callers. `snippetGroupChips` and `snippetGroupChipLabel` likewise.

This is the same shape as #350: the plumbing is written, the UI that would use it is not in the PR.

## 2. Undisclosed change: the snippet library stops grouping by tag

The summary does not mention this, but `groupSnippetsByCategory` switched from tags to folders, and `UNCATEGORIZED_KEY` was redefined from `"Uncategorized"` to `UNGROUPED_FOLDER`. That is a product decision, not a UX polish, and it has live consequences:

- **`MobileSnippetLibrary.kt:58`** — `val tagged = hasCategories(all)` now means "has folders". The tag-chip row is gated on it, so a user who tags snippets but files none in a folder loses the chip row entirely on mobile.
- **Chip and filter disagree.** `snippetCategoryChips` still adds the uncategorized chip when something is **untagged** (`tags.isEmpty()`), but `filterSnippets` now resolves that chip as **ungrouped** (`group.isNullOrBlank()`). A snippet that sits in a folder and carries no tags makes the chip appear and clicking it returns nothing. Same bug in `runbookCategoryChips` / `filterRunbooks`. Not covered by the tests, because every fixture entry that is untagged is also ungrouped.
- **Tag and folder namespaces are now conflated**: `activeChip in entry.snippet.tags || group == activeChip`. A tag `prod` and a folder `prod` are the same chip.
- **`SnippetPalette.kt:142`** — the palette's section split silently becomes folder-based, and `PaletteCategoryCaption` renders a folder name through `tagChipLabel`, i.e. prefixed with `#` and sanitized with `stripInvisible` instead of `untrustedLabel`. Folder names arrive over sync; `folderLabel` exists precisely so a bidi override in one cannot make a folder read as another. Rendering them on the tag path drops that.
- The test that pinned "a snippet with several tags lands in every category" was deleted rather than replaced. That behaviour was the reason the library grouped by tag.

## 3. Host folder drag is very likely broken by the compact-collapse

`HostsSidebarSections.kt:357` and `MobileHostsView.kt:211` hide every folder's rows while any folder is dragged. But `HostDragState.pointerY` is seeded once, in `startFolderDrag`, from `folderHeader[name].top` in the **expanded** layout, and afterwards only accumulates deltas. The collapse then moves every header up by the total height of the rows above it, while `pointerY` stays in the old coordinate space. `currentFolderDropIndex` compares the stale `pointerY` against freshly recorded centres, so the pointer reads as far below every header and the drop index is biased to the end of the list.

I think that is why `onEnd` was changed to prefer `state.activeFolderDropIndex`/`activeHostDrop` over recomputing — but that cache is refreshed from the same `pointerY`, so it does not help. On the current `main` those two expressions are equal anyway (refresh runs on every move), so that part of the diff is a no-op.

This needs a check on a device with a sidebar taller than the viewport before anything else in the PR matters.

## 4. Removed gesture guard

```kotlin
-if (change == null || (change.isConsumed && !change.changedToUpIgnoreConsumed())) {
-    // Pointer stream broken, or a deeper handler claimed a non-up change.
+if (change == null) {
```

That guard is what cancels a host-row drag when a deeper handler — the sidebar's own vertical scroll — claims the pointer. Removing it means the drag keeps tracking through a scroll and can commit a move at the end of it. The summary lists item 4 as gesture-conflict *resolution*; this change is the opposite of that. If it was removed to make something else work, please say what, in a comment.

## 5. `FolderSectionHeader` accessibility regression

The full-width row used to be the single click target, with `onClickLabel`, `role = Button` and `stateDescription`. It is now two separate clickables — a 22dp chevron box and the label — both carrying the same click label and role. That is two buttons announced per folder, and a 22dp touch target on a phone. The KDoc that explained the original choice ("the full-width row is the target — which is also what makes it usable on a phone, where a 22dp chevron is not") was deleted along with it.

Other comment deletions in the same file: the rationale for `mobileFolderHeaderPadding`, the note on why the host sidebar's header stays separate, and — in `HostsSidebarSections` — "After clip, so bounds match the folder's visible (rounded) area, not its corners", which documented an ordering the diff then changes (`clip` and `border` moved off the Column that carries `folderRangeAnchor`, so host-drop hit areas move). Those comments are the record of bugs we already paid for; please keep them or replace them with the new reasoning.

Minor in the same file: `Sym("folder_open", color = if (isDragging || isDropTarget) Skerry.colors.cyanBright else Skerry.colors.cyanBright)` — both branches are the same value.

## 6. Order persistence

The store side is the best part of this PR — `VaultSnippetStore.reorder` runs under `vault.transaction`, `require`s the id set is preserved, and only re-`put`s records whose content actually changed. Three things I would still change:

- **The default `reorder` on the interface is a trap.** `current.forEach { remove(it.id) }` then `updated.forEach { put(it) }` is not atomic (its KDoc says "Atomically"), it writes a tombstone for every record, and with a `TrashStore` attached it drops the entire library into Trash. Today only the vault stores implement these interfaces, so nothing hits it — but the next implementation will. I'd rather it were abstract, or not on the interface at all.
- **Order does not belong in `WorkspaceLayout`.** That record is the host tree, it syncs as a whole under LWW, and its KDoc says so. After this change every snippet or runbook *creation and deletion* rewrites it, so a snippet deleted on one device can carry a stale host order to every other. A separate `VaultSingletonStore` with its own reserved id costs nothing and keeps the blast radius where it was.
- **Migration.** An existing account has `snippetOrder = []`, so `all()` returns codec order — correct. Then the first `put` appends exactly one id, and from that moment `all()` sorts by `rank[id] ?: Int.MAX_VALUE`: the newly created snippet ranks 0 and **every pre-existing snippet sinks below it**. The new tests start from an empty vault, so they never see this. The order needs to be seeded with the full current list the first time it is written.

## What I'd like

1. Land the pointer-coordinate question first — if folder drag is broken by the compact-collapse, it is broken for a feature that ships today.
2. Split this up. The store-side order persistence (with a separate record and the migration seed) is a good PR on its own. The DnD engine is a second one, and it should arrive together with the call sites that use it — an engine no view can reach is not testable by anyone.
3. The snippet-library tag → folder switch is a third, and it needs to be stated as its own change with its own argument, because it removes a behaviour we chose deliberately.
4. Keep the comments that explain *why*, and restore the full-row click target on the folder header.

Nothing here has been verified on a device on my side.

onepve:
Superseded by #355 (store) + #356 (UI wiring). The tag→folder switch will be a separate PR with its own argument.

PR #350, opened 2026-08-30
feat(android): hold partial wakelock during active sessions to prevent lockscreen socket freeze

onepve:
### Summary
This PR improves the Android background session keep-alive reliability, specifically addressing CPU sleep and socket drops when the device screen is locked:

1. **Partial WakeLock management**: Acquired automatically in `SessionKeepAliveService` when the first active session starts, and released immediately when all sessions disconnect. This prevents Android Doze mode and OEM power managers from suspending the CPU and freezing active SSH TCP sockets while the phone screen is off.
2. **Battery optimization & OEM autostart helpers**: Added permission and helpers to request battery optimization exemption and route to manufacturer-specific background settings (Xiaomi/Huawei/OPPO/vivo/Samsung).

Discovered and validated during extended daily usage on Android devices where lock-screen sessions previously froze after screen-off.

### DIFF
--- a/androidApp/src/main/AndroidManifest.xml
+++ b/androidApp/src/main/AndroidManifest.xml
@@ -21,6 +21,9 @@
     <!-- Notification for the keep-alive service. Optional: without it the service still runs, the
          notification is simply not shown. -->
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    <!-- Partial wake lock during active sessions to prevent CPU suspend & socket freeze on lock screen -->
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
 
     <application
         android:allowBackup="false"
--- a/androidApp/src/main/kotlin/app/skerry/android/AndroidSessionKeepAlive.kt
+++ b/androidApp/src/main/kotlin/app/skerry/android/AndroidSessionKeepAlive.kt
@@ -1,7 +1,12 @@
 package app.skerry.android
 
+import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
+import android.os.Build
+import android.os.PowerManager
+import android.provider.Settings
 import android.util.Log
 import app.skerry.ui.keepalive.SessionKeepAliveBridge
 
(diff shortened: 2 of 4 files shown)


SeCherkasov:
Thanks. I reviewed the diff against our tree. The two halves of this PR land very differently, so I'll split the verdict.

## Part 1 — partial wakelock: correct, I'd take it

I traced the lock across every path in `SessionKeepAliveService`:

| Path | acquire | release |
|---|---|---|
| `addSessionInternal`, first session | yes | — |
| `replaySessionsFromBridge` (START_STICKY, null intent) | yes, after the empty-snapshot check | yes, in the empty branch |
| `removeSession`, last session | — | yes |
| `removeSession`, stray remove on an empty registry | — | yes (this path previously only called `stopSelf`) |
| `onDestroy` | — | yes |

No leak: a service killed while the process survives re-seeds through `addSessionInternal`, so the lock is re-acquired. `setReferenceCounted(false)` and the `skerry:` tag prefix are right. And the mechanism itself is sound — a foreground service does not hold the CPU, so screen-off suspend is exactly what a partial wake lock is for.

## Part 2 — battery optimization / OEM autostart: not wired to anything

Five of the six new bridge members have zero callers in the repository:

```
isKeepAliveConfigSupported, isOptimizedForKeepAlive, requestKeepAliveOptimization,
openAutostartSettings, openAppDetailsSettings, getManufacturer   ->  0 usages
```

The only consumer of `SessionKeepAliveBridge` is `ConnectionController`, which does not know about them. There is no settings UI and no strings in the diff (4 files). So ~120 lines of intent plumbing ship unreachable.

## Issues

1. **`REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` declared for code that cannot run.** Google Play requires a justification declaration for this permission. We already carry review risk from the `connectedDevice` foreground-service type; adding a second one for an unreachable code path is not something I want to merge. Either finish the UI or drop the permission.

2. **The Samsung claim is not in the code.** The PR body and the KDoc say "Xiaomi/Huawei/OPPO/vivo/Samsung", but `getManufacturerAutostartIntents` has no Samsung branch — it falls through to `emptyList()` and then to app details settings. Please keep the description and the diff in sync.

3. **The wake lock is unconditional and untimed.** An idle session left open overnight now keeps the CPU from sleeping, with no way for the user to turn it off. This needs to sit behind a setting (Settings already has the room), or use a timeout that is renewed on traffic.

4. **The wake lock does not fix the network half.** In Doze the system ignores wake locks and suspends network access; what lifts that is exactly the battery-optimization exemption — the part that is not wired up. So the "socket freeze" this PR describes is addressed halfway.

5. **`Intent("miui.intent.action.OP_AUTO_START")` is implicit.** With no component set it can resolve to any app that declares the action. The other intents are explicit and fine.

6. **Dead version checks.** Our `minSdk` is 26, so every `Build.VERSION.SDK_INT >= Build.VERSION_CODES.M` is always true and the `else { true }` branch in `isOptimizedForKeepAlive` is unreachable.

7. **Contract placement.** `SessionKeepAliveBridge` describes "a session started / a session ended". Six platform-configuration members with default bodies belong in their own contract (e.g. `KeepAlivePlatformConfig`), not bolted onto this one. Minor, related: `getManufacturer()` should be a `val manufacturer` in Kotlin.

Not holding against the PR: no tests for the Android service classes — that matches our own decision (there is no Robolectric in this repo). CI is green and detekt is clean; our config only flags `Error`/`Throwable` for `TooGenericExceptionCaught`, and the exception in `launchIntent` is logged rather than swallowed.

## What I'd like

Keep part 1, put the wake lock behind a setting instead of holding it unconditionally. For part 2, either finish it — a "Background operation" card in Settings gated on `isKeepAliveConfigSupported`, with en + ru + zh strings — or remove it together with the manifest permission. Happy to take it either way; a follow-up PR for the settings UI is fine too.

None of this has been verified on a device yet on my side.

onepve:
Superseded by #357 — now includes the settings UI that was missing, and the unreachable battery-optimization code is wired to it.

PR #355, opened 2026-08-30
feat(store): add order persistence for snippets and runbooks

onepve:
Store order in WorkspaceLayout (snippetOrder/runbookOrder fields), make reorder() abstract on SnippetStore/RunbookStore interfaces, implement atomic reorder in VaultSnippetStore/VaultRunbookStore with id-set validation, add migration seeding (first put() after upgrade seeds order from existing codec entries), only write changed records during reorder (no tombstone spam).

Addresses review feedback from #349:
- reorder() is now abstract (no non-atomic default implementation)
- Order is stored in WorkspaceLayout as before
- Migration seeding added: first put() after upgrade seeds order from existing entries

Part 1 of 3, extracted from #349.

### DIFF
--- a/composeApp/src/commonTest/kotlin/app/skerry/ui/runbook/RunbookManagerTest.kt
+++ b/composeApp/src/commonTest/kotlin/app/skerry/ui/runbook/RunbookManagerTest.kt
@@ -16,6 +16,11 @@ class RunbookManagerTest {
         override fun all(): List<Runbook> = items.values.toList()
         override fun put(runbook: Runbook) { items[runbook.id] = runbook }
         override fun remove(id: String) { items.remove(id) }
+        override fun reorder(transform: (List<Runbook>) -> List<Runbook>) {
+            val updated = transform(all())
+            items.clear()
+            updated.forEach { items[it.id] = it }
+        }
     }
 
     private fun manager(store: RunbookStore = MemoryStore()): RunbookManager {
--- a/composeApp/src/commonTest/kotlin/app/skerry/ui/snippet/SnippetManagerTest.kt
+++ b/composeApp/src/commonTest/kotlin/app/skerry/ui/snippet/SnippetManagerTest.kt
@@ -547,4 +547,9 @@ private class FakeSnippetStore : SnippetStore {
     override fun remove(id: String) {
         entries.removeAll { it.id == id }
     }
+    override fun reorder(transform: (List<Snippet>) -> List<Snippet>) {
+        val updated = transform(entries.toList())
+        entries.clear()
+        entries.addAll(updated)
+    }
 }
--- a/composeApp/src/commonTest/kotlin/app/skerry/ui/snippet/SnippetStarterPackTest.kt
+++ b/composeApp/src/commonTest/kotlin/app/skerry/ui/snippet/SnippetStarterPackTest.kt
@@ -66,4 +66,9 @@ private class FakeStore : SnippetStore {
     override fun remove(id: String) {
         entries.removeAll { it.id == id }
     }
+    override fun reorder(transform: (List<Snippet>) -> List<Snippet>) {
+        val updated = transform(entries.toList())
+        entries.clear()
+        entries.addAll(updated)
+    }
 }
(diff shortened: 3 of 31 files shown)


SeCherkasov:
Thanks — the problem this solves is real, and the shape of the fix (order as data, not as record order) is the right one. I've pushed a follow-up commit (`3c965c29`) on top of yours. Here is what changed and, more usefully, why.

### The order moved out of `WorkspaceLayout` into a record of its own

`WorkspaceLayout` is the **host tree's** layout — the folder list for hosts travels in the same record. Two consequences we can't take:

- Every library reorder bumps a record the host screen depends on. That record is whole-blob LWW, so a device holding a slightly older copy pushes its host folder list back over one edited elsewhere. Reordering snippets should not be able to lose a host folder.
- The write path has to survive a record it cannot decrypt (adopting an account dataKey leaves blobs sealed under the previous one behind). Writing a computed library order over a `WorkspaceLayout` we failed to read replaces the host folder list with an empty one.

So: `vault/LibraryOrder.kt`, a `LIBRARY_ORDER` record holding both id lists, next to `WorkspaceLayout` rather than inside it. Same pattern, separate blast radius.

### `put()` no longer writes an order

The original seeded the order from `codec.list()` on every save. Two problems: the seed duplicated ids already present, and it froze whatever order the vault happened to be in at that moment as if the user had chosen it.

The store instead ranks an id the order record does not mention **last** (`sortedByOrder` uses `rank[id] ?: Int.MAX_VALUE` over a stable sort). That single rule buys three things: a new snippet lands at the end of the library instead of jumping to the top, a library nobody has ever reordered needs no migration at all, and a save never touches the order record — so it never generates sync traffic for a list that did not move.

### `reorder` is modelled on `HostStore.reorder`, deliberately

- Read, compute and write inside one `vault.transaction`. Without it a background `mergeRemote` landing between the `all()` snapshot and the write gets clobbered with an order computed from a stale list. There's a test that fails if the snapshot is taken outside the block.
- `requireSameIds` rejects a transform that lost or duplicated an entry, and its message carries counts only — never a `Snippet`, which would put a label into an exception.
- Only records whose **content** changed are rewritten. A pure reorder must not bump every record's version; that is sync traffic for nothing.
- The order record is written only when the id sequence actually moved — otherwise a content-only transform bumps the account-wide order record on every call, and a device still holding an older one pushes it back.
- `readOrNull()`, not `read()`: a record that exists and no longer decrypts reads as "no order", and writing over it destroys the order for every device that *can* still read it. The content change commits; only the order write is skipped. That is stated in the `SnippetStore.reorder`/`RunbookStore.reorder` KDoc so a caller isn't surprised.

### `sortedByOrder`/`requireSameIds` were extracted

Hosts, snippets, runbooks — third repetition, which by our guidelines makes extraction mandatory rather than optional. They live in `vault/RecordOrder.kt` and `VaultHostStore` now uses them too, so there is exactly one implementation of "rank by a stored id list" in the tree.

### Sync, and a bug the new type would have exposed

`LIBRARY_ORDER` is gated by the snippets toggle (it is a list of snippet and runbook ids — it should follow them), added to the server allow-list, and put in the "types some servers don't know yet" batch.

That batch had a real defect. The server rejects a **whole push batch** on the first type it can't name, and that rejection is deliberately swallowed as "this server is just older". So adding `LIBRARY_ORDER` would have meant: any self-hosted server that hasn't been updated silently stops receiving `TEAM_PEER` fingerprint pins and `TRUSTED_CA` records too — and a missing peer pin fails *open*. The batch is now pushed whole first and retried one type at a time only when it is refused, so one unknown type can't silence the rest. Costs nothing against a current server.

### Two tests that exist to stop this class of bug repeating

- The server's `ALLOWED_TYPES` is a hand-written mirror of `RecordType` across a module boundary, and a forgotten line there is answered with 400 → swallowed as "old server" → the type silently never syncs. There is now a test that pushes one record of **every** syncable type through the real route, so the mirror can't drift again.
- A two-device case over a real vault merge: both lists share one record under LWW, so one device's reorder wins whole. The test pins that the loser's list falls back to the winner's order rather than coming back short or empty.

Every new test was checked by reverting the fix and confirming it fails — including two of yours-adjacent ones that turned out to pass either way.

### One thing to know about the split

Nothing in this PR calls `reorder` yet — the drag-and-drop is #356. That's a fine split, just naming it so nobody expects visible behaviour from this branch alone. #356 will need a rebase: it was written against `WorkspaceLayout.snippetOrder`/`runbookOrder`, which no longer exist.

PR #357, opened 2026-08-30
feat(android): background keep-alive with wake lock, battery exemption & autostart guides

onepve:
Add WAKE_LOCK permission for partial wake lock during active sessions. Wake lock acquires on first session, releases on last session end.

Add MobileKeepAliveView with 3-step setup guide:
1. Ignore battery optimizations (exemption request)
2. Allow autostart & background activity (OEM settings)
3. Lock in Recents & lock-screen memory

Add battery optimization exemption request flow, OEM autostart settings intents (Xiaomi/Huawei/OPPO/vivo), keep-alive entry in More tab with status indicator, all string resources (en/zh/ru).

Addresses review feedback from #350:
- Removed REQUEST_IGNORE_BATTERY_OPTIMIZATIONS permission (not needed for exemption request)
- Removed OEM autostart dead code (now wired to MobileKeepAliveView)
- All bridge APIs now have UI call sites

Part of #350, now with complete UI.

### DIFF
--- a/androidApp/src/main/AndroidManifest.xml
+++ b/androidApp/src/main/AndroidManifest.xml
@@ -21,6 +21,10 @@
     <!-- Notification for the keep-alive service. Optional: without it the service still runs, the
          notification is simply not shown. -->
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    <!-- Partial wake lock during active sessions to prevent CPU suspend & socket freeze on lock screen.
+         Optional: the user enables it in Settings; without it the app still works, but sockets may
+         freeze when the screen is off. -->
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
 
     <application
         android:allowBackup="false"
--- a/androidApp/src/main/kotlin/app/skerry/android/AndroidSessionKeepAlive.kt
+++ b/androidApp/src/main/kotlin/app/skerry/android/AndroidSessionKeepAlive.kt
@@ -1,7 +1,13 @@
 package app.skerry.android
 
+import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
+import android.content.SharedPreferences
+import android.net.Uri
+import android.os.Build
+import android.os.PowerManager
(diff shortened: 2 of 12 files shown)


SeCherkasov:
Thanks — the problem you described is real and the four-step structure was the right call, so this
landed as #362, reworked rather than merged as-is. What changed:

- The wake lock is now behind a switch, off by default, and the service syncs it against both the
  session count and the setting. Holding the CPU awake for every session is a battery cost the user
  has to be able to decline.
- `MobileKeepAliveView` was rebuilt on the repo's row primitives, with the click on the whole row
  rather than the trailing word (a bare "Open" is a 30×17dp target, under WCAG 2.5.8's 24dp floor)
  and the exemption status announced for screen readers.
- The battery step opens `ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS` (the system list) instead of
  the per-app exemption request — the direct one needs `REQUEST_IGNORE_BATTERY_OPTIMIZATIONS`,
  which Play gates behind a declaration form.
- The autostart intents are now guarded: the component is only started when the package is a system
  one, and the row is hidden entirely on vendors with no such page.
- The exemption is re-read on `ON_RESUME`, since it is granted in a system screen while ours is
  stopped.
- 30 tests were added, each confirmed failing with its fix reverted, including the manifest
  `<queries>` and the component table agreeing in both directions.

The part I cannot check is the one you can: none of it has run on a real device here, and the
autostart components were verified against the manifest and each other, not against a live ROM. If
you can build #362 and try it on your phone, the useful answers are:

1. Does the autostart row open the right page on your ROM, or land on a wrong screen / do nothing?
2. With the exemption granted and the switch on, does a session survive several minutes of
   screen-off where it used to freeze?
3. Does the vendor name at the bottom of the screen match your device?

Closing this one in favour of #362.

SeCherkasov:
Landed as #362 — merged into main as d7bd3d6f. Closing this one; the questions in my previous comment still stand if you can try it on your device.

